PR #28, opened 2020-01-31
Release/0.1.0 alpha.15

colmsnowplow:
- Updated tracker to use ios-1.2.0
- Updated DemoApp to latest and greatest version
- Updated DemoApp's tracking instrumentation

No huge rush on this one guys, I'll release final asset, cleanup and rebase after I'm back from the few days off. If you are curious & want to test, the 0.1.0-alpha.M2 asset is available on npm.


istreeter:
I think you need to git-add the file DemoApp/android/gradle/wrapper/gradle-wrapper.jar to the repo. I couldn't build the demo app without it, 

From the [gradle docs](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.wrapper.Wrapper.html):

> The scripts generated by this task are intended to be committed to your version control system. This task also generates a small gradle-wrapper.jar bootstrap JAR file and properties file which should also be committed to your VCS. The scripts delegates to this JAR.

(if you don't want to add the jar, then you could add instructions to the readme how to generate the gradle-wrapper jar manually)

DemoApp/App.js
```diff
@@ -6,57 +6,124 @@
...
-            title="Send event"
-            color="#841584"
-            accessibilityLabel="Send an event"
-        />
-      </View>
-    );
+const App: () => React$Node = () => {
+  Tracker.initialize('test-endpoint', 'post', 'https', 'namespace', 'app-id', {autoScreenView: false});
+  Tracker.trackScreenViewEvent('Name', null, null, null, null, null, null, []);
+  let onPressSendEvent = () => {
```
istreeter: or const instead of let

DemoApp/.flowconfig
```diff
@@ -5,19 +5,18 @@
 ; Ignore "BUCK" generated dirs
```
istreeter, line 5: Flow+react is a great combination.  You could add flow-bin as a devDependency in package.json.  Mainly so that all developers are using the same version of flow.  Also consider adding it as a script in package.json.

DemoApp/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.5-all.zip
```
istreeter: FWIW the latest version of gradle is 6.1.1

istreeter (commented):
I confirm the Android demo app worked for me.  I think I cannot test the ios version on my os.

paulboocock (commented):
This works on iOS in my testing. However, I've noticed that we've included the Pods directories. They don't need commiting and should be added to `.gitignore`. A pod install should pull all the dependencies (in fact, I couldn't get it to build the iOS demo app until I deleted the Pods folder and reinstalled).

DemoApp/android/gradle/wrapper/gradle-wrapper.properties
colmsnowplow: Updated

DemoApp/.flowconfig
colmsnowplow: Added to dev dependencies

colmsnowplow:
Thanks for the review @paulboocock @istreeter.

Think I've addressed all the comments. Apologies for the horrendous amount of files changed - this is due to removing the Pods folders.

If you're happy to, I'll rebase & release tomorrow :)

paulboocock (approved):
LGTM

PR #35, opened 2020-04-02
ISSUE-34-1: Add Session Context

max-tgam:
Allow configuring the session context by passing following optional values via options object.
Available options:

- `setSessionContext` (boolean), defaults to `false`, if set to `true` allows the following parameters
- `checkInterval` (integer), optional, only works if `setSessionContext` is set to `true`, defaults to `15` (seconds)
- `foregroundTimeout` (integer), optional, only works if `setSessionContext` is set to `true`, defaults to `600` (seconds)
- `backgroundTimeout` (integer), optional, only works if `setSessionContext` is set to `true`, defaults to `300` (seconds)


colmsnowplow:
Cherry picked into #49

PR #36, opened 2020-04-02
ISSUE-34-2: Add Application Context

max-tgam:
Allow enabling application context by passing following optional values via options object.
Available options:
- `setApplicationContext` (boolean), defaults to `false`


colmsnowplow:
Cherry picked into #49

PR #37, opened 2020-04-02
ISSUE-34-3: Add Life Cycle Events

max-tgam:
Allow enabling lifecycle events passing following optional values via options object.
Available options:
- `setLifecycleEvents` (boolean), defaults to `false`


colmsnowplow:
Cherry picked into #49

PR #39, opened 2020-04-02
ISSUE-34-5: Allow Base64 configuration

max-tgam:
Allow configuring base64 encoding by passing following optional values via options object.
Available options:

- `setBase64Encoded` (boolean), defaults to `false`


colmsnowplow:
Cherry picked into #49

PR #40, opened 2020-04-02
ISSUE-34-6: Allow Install and Exception Event config

max-tgam:
Allow configuring Install and Exception Events by passing following optional values via options object. The Install event is working on iOS, on Android it needs additional instrumentation inside React Native App. Exception tracking needs additional instrumentation inside React Native App.
Available options: 

- `setInstallEvent` (boolean), defaults to `false`
- `setExceptionEvents` (boolean), defaults to `false`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,10 @@ build/
 .gradle
 local.properties
 *.iml
+android/.project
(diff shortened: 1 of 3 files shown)


colmsnowplow:
Cherry picked into #49

PR #41, opened 2020-04-02
ISSUE-34-7: Allow enabling Platform and Geo contexts

max-tgam:
Allow enabling/disabling Platform and Geo contexts by passing following optional values via options object. Geo contexts might need additional instrumentation inside React Native App.
Available options:

 - `setPlatformContext` (boolean), defaults to `false` (*NOTE:* this is different from current ssettings!)
- `setGeoLocationContext` (boolean), defaults to `false`


colmsnowplow:
Cherry picked into #49

PR #49, opened 2020-04-09
Release/0.1.0 alpha.16

colmsnowplow:


colmsnowplow:
@elliotstokes Finally ready, sorry for the back and forth. 

One final question - I haven't committed the updated package.lock, yarn.lock, or Podfile.lock files, like I was doing before. I have a hunch that these can just be removed. 

Do you know what best practice is regarding those kinds of files for react-native projects?

elliotstokes:
> @elliotstokes Finally ready, sorry for the back and forth.
> 
> One final question - I haven't committed the updated package.lock, yarn.lock, or Podfile.lock files, like I was doing before. I have a hunch that these can just be removed.
> 
> Do you know what best practice is regarding those kinds of files for react-native projects?

I guess it depends but if the lock files are there it'll stop dependencies updating even if you've specified something less restrictive in the pod file or the package.json. I think it's probably fine in this case to leave them out.

elliotstokes (approved):
LGTM

brianchitester:
Hello, thanks for your hard work making this package available!  I just wanted to chime in on your lockfile question since I am planning to use this for a project I'm working on.

First, you should only have either package-lock.json, or yarn.lock in your project. package-lock.json is used by npm, and yarn.lock is used by yarn.  You will need to use one or the other consistently or you will run into problems.

Second, lockfiles should always be committed. Yarn's website has a detailed description as to why this is important: https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

I hope you find this helpful and thanks again!

colmsnowplow:
Thanks for the input - the reason the question came up in the first place was that I noticed both lock files were present, and the demoApp is primarily a testing tool, and mostly those tests require a full rebuild & re-import of at least the asset we're testing, I was mulling over whether to include either. (You might notice the script in this pr just deletes all of those files.

I'll leave yarn.lock in, for the sake of anyone who might want to use the demoApp as an easy way to play with the tracker.

PR #52, opened 2020-04-29
Update examples for timezone and IP address

miike:
- Replace deprecated EST time zone key with a different canonical zone name
- Replace IP address with valid IP in case users test with a non-valid IP


colmsnowplow (approved):
Good spot, thanks for cleaning that up @miike. Merging. Hope all's well with you.

PR #56, opened 2020-05-04
Release/0.1.0 alpha.17

colmsnowplow:


CHANGELOG
```diff
@@ -1,13 +1,18 @@
+Version 0.1.0-alpha.167 (2020-05-04)
```
paulboocock: Typo, should be alpha.17

paulboocock (approved):
LGTM. I’ll let @AlexBenny comment on the FMDB version updates with iOS 1.3.0 - I think the updates here are correct but I recall some issue in 1.3.0 with FMDB so worth a second pair of eyes.

CHANGELOG
colmsnowplow: Ah! Thanks.

AlexBenny (approved):
I haven't tested it on my machine but it looks good to me! 🚢 
About the issue with FMDB, it's mostly related to Swift Package Manager so it's not an issue on ReactNative, however it will be released soon a patch version 1.3.1 of the iOS tracker.

PR #68, opened 2020-06-05
Method instrumentation

colmsnowplow:


DemoApp/App.js
```diff
@@ -28,13 +28,21 @@ import {
...
 const App: () => React$Node = () => {
-  Tracker.initialize('test-endpoint', 'post', 'https', 'namespace', 'my-app-id', {
+  Tracker.initialize({
+    // required
+    endpoint: 'test-endpoint',
+    method: 'post',
+    protocol: 'https',
+    namespace: 'namespace',
+    appId: 'my-app-id',
+
```
AlexBenny, line 38: As we discussed in our call, it would be nice to have config objects in the future:
- one for endpoint params, where `endpoint` field is mandatory in the _constructor_ and `method` and `protocol` can be optional with default values.
- one for other optional params
Not sure where `namespace` and `appId` can go.

index.js
```diff
@@ -1,6 +1,108 @@
...
+
+    let defs = {
+      property: null,
+      label: null
+    }
+
+    // RN's Java bridge forces value into double, which throws on null or undefined.
+    // Remove 'value' key if null or undefined to work around this limitation
+    if (argmap['value'] == null) {
+      delete(argmap['value'])
```
AlexBenny: Pitiful that RN haven't fixed this issue yet.
I wonder if it can be generalised to filter out all the types with a primitive counterpart.

AlexBenny (approved):
Great job! Just added minor comments for reference that we already discussed in our call.

android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
```diff
@@ -120,34 +135,79 @@ public void trackSelfDescribingEvent(ReadableMap event, ReadableArray contexts)
...
+        )) {
+         promise.reject("ERROR", "SnowplowTracker: trackStructuredEvent() method - missing parameter with no default found");
+        }
+
+        Structured trackerEvent = EventUtil.getStructuredEvent(
+          details.getString("category"),
+          details.getString("action"),
+          details.getString("label"),
+          details.getString("property"),
+          details.hasKey("value") ? details.getDouble("value") : null,
```
AlexBenny: If you check with `details.isNull("value") ? ....` maybe it's not needed to remove the "value" param from the args.
Just wondering!

colmsnowplow: Good point. Missed that completely. 

Ian's input in slack has made me question whether the pattern should change anyway though - setting null defaults to work around limitations of methods seems to abstract the solution away from the problem. 

Fixing the util methods, and conditionally removing builder arguments may be the preferable approach... I'm mulling it over!

colmsnowplow:
@istreeter @AlexBenny - following our discussion, I have made some changes to the refactor here: https://github.com/snowplow-incubator/snowplow-react-native-tracker/pull/68/commits/49b94007884588abdcfd9cac0f177d7e2737450f

I think I prefer this pattern for two reasons:

1. Specifically handling a problem for a given platform puts the solution next to the problem (eg where iOS doesn't like a null, that's handled where the nullable value is called, within the iOS code).
2. Instead of setting things as null or undefined, we're not including the builder argument to set that thing - IMO this is preferable as it's more like what you'd do if implementing the tracker in a native app.

Any thoughts let me know!

istreeter:
I like this new pattern, for exactly the same two reasons.

colmsnowplow:
Have cherry-picked relevant commits into develop. Closing.

PR #76, opened 2020-06-11
Release/0.1.0

colmsnowplow:


DemoApp/App.js
```diff
@@ -47,7 +55,7 @@ const App: () => React$Node = () => {
       screenHeight: 456,
       colorDepth: 20,
       timezone: 'Europe/London',
-      language: 'en',
+      language: 'fr',
```
AlexBenny, line 58: 🥖

DemoApp/App.js
```diff
@@ -56,11 +64,52 @@ const App: () => React$Node = () => {
...
+  }
+  const onPresstrackScreenViewEvent = () => {
+    Tracker.trackScreenViewEvent({screenName: 'onlyRequired'});
+    Tracker.trackScreenViewEvent({screenName: 'allPopulated', screenType: 'allPopulated', transitionType: 'test' });
+    Tracker.trackScreenViewEvent({screenName: 'allOptionalsNull', screenType: null, transitionType: null});
+    Tracker.trackScreenViewEvent({screenName: 'allOptionalsUndefined', screenType: undefined,transitionType: undefined});
+    Tracker.trackScreenViewEvent({screenName: 'withAContext'}, [{schema: "iglu:com.snowplowanalytics.snowplow/gdpr/jsonschema/1-0-0", data: {basisForProcessing: "consent"}}]);
+    Tracker.trackScreenViewEvent({screenName: 'withEmptyArrayContext'}, []);
+  }
+  const onPresstrackSelfDescribingEvent = () => {
```
AlexBenny: typo:
```suggestion
  const onPressTrackSelfDescribingEvent = () => {
```

README.md
```diff
@@ -44,36 +47,40 @@ Tracker.trackSelfDescribingEvent({'schema': 'iglu:com.acme/hello_world_event/jso
...
+initialize({
+  // required
+  endpoint: 'my-endpoint',
+  method: 'post',
+  protocol: 'https',
+  namespace: 'my-namespace',
+  appId: 'my-app-id',
+
+  // optional
+  setPlatformContext: true,
```
AlexBenny: Minor doubt about naming convention:
The prefix `set` is usually used on methods. I wonder if it would be more appropriate just `platformContext` or `platformContextEnabled`, considering we are setting fields of a configuration object.

index.js
```diff
@@ -1,6 +1,85 @@
...
+    }
+  }
+
+  static setSubjectData(argmap) {
+    RNSnowplowTracker.setSubjectData(argmap);
+  }
+
+  static trackScreenViewEvent(argmap, ctxt=[]) {
+
+    //TODO: Remove previousScreenName, previousScreenType, previousScreenId - shouldn't be settable and don't appear to work on iOS.
```
AlexBenny: 👍

ios/RCTConvert+Snowplow.m
```diff
@@ -5,52 +5,52 @@
...
-    	}
-    	SPSelfDescribingJson *sdj = [[SPSelfDescribingJson alloc] initWithSchema:schema andData:data];
-    	return sdj;
-	}
-	return nil;
+    				return nil;
+	    	}
+	    	SPSelfDescribingJson *sdj = [[SPSelfDescribingJson alloc] initWithSchema:schema andData:data];
+	    	return sdj;
+		}
```
AlexBenny: It seems tabs screwed up here on GitHub

ios/RCTConvert+Snowplow.m
```diff
@@ -5,52 +5,52 @@
...
+		        if (contexts != nil) {
+		            [builder setContexts:[[NSMutableArray alloc] initWithArray:contexts]];
+		        }
+		    }];
+		    return unstructEvent;
+	  } else {
+		    RCTLogConvertError(json, @"needs to be an NSDictionary with at least schema and some data type");
+		    return nil;
+	  }
+	  return nil;
```
AlexBenny: I think this return is never reached.
You could eventually use the fast fail here too:
```
if (![json isKindOfClass:[NSDictionary class]]) {
    RCTLogConvertError(json, @"needs to be an NSDictionary with at least schema and some data type");
    return nil;
}
SPSelfDescribingJson * data = [self SPSelfDescribingJson:json[@"event"]];
...
```

ios/RNSnowplowTracker.m
```diff
@@ -14,175 +14,216 @@ - (dispatch_queue_t)methodQueue
...
+                                :rejecter:(RCTPromiseRejectBlock)reject) {
+
+    // ios throws if it finds an NSNull here - unset the field if null passed
+    if (options[@"userId"] != nil) {
+        if (options[@"userId"] == (id)[NSNull null]) {
+            [self.tracker.subject setUserId:nil];
+        } else {
+            [self.tracker.subject setUserId:options[@"userId"]];
+        }
+    }
```
AlexBenny: All this could be also written as:
```suggestion
    NSString *userId = options[@"userId"];  // to limit number of access to dictionary
    if (userId) {
        NSString *newUserId = userId == [NSNull null] ? nil : userId;  // No cast to (id)
        [self.tracker.subject setUserId:newUserId];
    }
```

ios/RNSnowplowTracker.m
```diff
@@ -14,175 +14,216 @@ - (dispatch_queue_t)methodQueue
...
+            [self.tracker.subject setDomainUserId:nil];
+        } else {
+            [self.tracker.subject setDomainUserId:options[@"domainUserId"]];
+        }
+    }
+
+    if (options[@"screenWidth"] != nil && options[@"screenHeight"] != nil) {
+        if (options[@"screenWidth"] == (id)[NSNull null] || options[@"screenHeight"] == (id)[NSNull null]) {
+            NSError * error = [NSError errorWithDomain:@"SnowplowTracker" code:100 userInfo:nil];
+            reject(@"ERROR", @"SnowplowTracker: initiasetSubjectDatalize() method -  screenWidth and screenHeight cannot be null", error);
```
AlexBenny: I'm in doubt. Does `reject` stop the execution of the method too?
If no, does it make sense to return after reject?

ios/RNSnowplowTracker.m
```diff
@@ -14,175 +14,216 @@ - (dispatch_queue_t)methodQueue
...
-                  :(NSString *)screenId
-                  :(NSString *)screenType
-                  :(NSString *)previousScreenName
-                  :(NSString *)previousScreenType
-                  :(NSString *)previousScreenId
-                  :(NSString *)transitionType
+                  :(NSDictionary *)details
                   :(NSArray<SPSelfDescribingJson *> *)contexts) {
+
     SPScreenView * SVevent = [SPScreenView build:^(id<SPScreenViewBuilder> builder) {
```
AlexBenny: variable better lowercased.
On ObjC the convention is long self-explaining names so it would be perfectly fine `screenViewEvent`

AlexBenny (approved):
Giant step forward! I've just put some notes, not a blocker.
I've tried to launch the demo apps but when I've launched Android my macOS crashed. The iOS app started and worked well!
I've used `cleanBuild.sh` that actually start the apps too, so maybe we could rename it.
I wonder if it would be nice to indicate that script in the README file.

ios/RNSnowplowTracker.m
colmsnowplow: reject returns a rejected promise, and yes stops execution.

colmsnowplow:
@AlexBenny your entire machine crashed on running android??

DemoApp/App.js
```diff
@@ -56,11 +64,52 @@ const App: () => React$Node = () => {
...
 
-  Tracker.trackScreenViewEvent('Name', null, null, null, null, null, null, []);
-  const onPressSendEvent = () => {
-    Tracker.trackSelfDescribingEvent({'schema': 'iglu:com.acme/event/jsonschema/1-0-0', 'data': {'message': 'hello world'}}, []);
-    Tracker.trackStructuredEvent('category', 'action', 'label', 'property', 50.00, []);
-    Tracker.trackPageViewEvent('acme.com', 'some title', 'refr.com', []);
+  initPromise.then(() => Tracker.trackScreenViewEvent({screenName: 'firstScreenView'}));
+
+  const onPresstrackStructuredEvent = () => {
+    Tracker.trackStructuredEvent({category: 'SeTest', action: 'allPopulated', label: 'valueIsFloat', property: 'property', value: 50.00});
```
istreeter, line 71: Should this be:

```
const onPresstrackStructuredEvent = () =>
  initPromise.then(() =>
    Tracker.trackStructuredEvent(..... // etc
```

Depending on how paranoid you want to be about the tracker not being initialized yet.

index.js
```diff
@@ -1,6 +1,85 @@
-
 import { NativeModules } from 'react-native';
 
 const { RNSnowplowTracker } = NativeModules;
 
-export default RNSnowplowTracker;
+function _applyDefaults(input, defaults) {
```
istreeter: I think this function is directly modifying the object passed in by the user, and it feels like that breaks a contract.  We shouldn't modify objects that don't "belong" to us.

I think this function can be rewritten as
```
function _applyDefaults(input, defaults) {
  return {...defaults, ...input}
```
(untested!)

index.js
```diff
@@ -1,6 +1,85 @@
...
+                 foregroundTimeout : 600,
+                 backgroundTimeout : 300,
+                 checkInterval : 15,
+                 setInstallEvent : false}
+
+
+    if (typeof argmap.endpoint !== 'undefined' && typeof argmap.appId !== 'undefined' && typeof argmap.namespace !== 'undefined' ) {
+      return RNSnowplowTracker.initialize(_applyDefaults(argmap, defs));
+    } else {
+      console.warn("SnowplowTracker: initialize() requires endpoint, namespace and appId parameter to be set")
```
istreeter, line 26: I always feel uneasy about console messages, because I feel they add to the bundle size of the package without doing anything helpful.

A clever developer can remove them from the production bundle anyway using webpack options, but I reckon most developers don't do this.

That's just my own thoughts though - I am not suggesting a change here.

index.js
colmsnowplow: Is it modifying the input? I understood it to copy the input to a new object and then return that?

I do like your way better though, much cleaner code. I'll give it a test.

index.js
colmsnowplow: So my options here felt like they were - throw an error (crashes app), throw a warning (popup appears in debug mode only), or have the native app crash.

Since a) the app crashing is a pain for development, and b) you don't really want an unexpected edge case to crash your entire app in production, this felt like the option to take.

Are there alternatives to consider here? 

PS. I could also return a promise rejection, but this has the same effect I believe?

index.js
colmsnowplow: Good spot. Your way is much better IMO. Thanks for flagging!

index.js
istreeter: You could try this:

```
} else if (__DEV__) {
  console.warn("you did something wrong")
}
```

With the implication that if not dev then nothing happens.  Just move on.

The interesting test is whether the react-native script is clever enough to remove this line entirely from a production build (e.g. of the DemoApp).  If it doesn't remove it, then this doesn't really achieve anything - so it would be fine to leave it as you have done it.

colmsnowplow: Ah yeah good point. It seems this may well do what we want it to do: https://stackoverflow.com/questions/55340673/is-there-a-way-to-keep-debug-code-out-of-a-release-build-in-react-native

Nice, thanks for the suggestion! I'll double check this and hopefully that'll be that nicely handled. :)

DemoApp/App.js
colmsnowplow: So, yes - it's a bit wonky the way I've done it, but I did it this way on purpose.

What I was aiming for was a test case which uses the promise, and a test case which doesn't, under the assumption that some won't use this structure.

Thinking about that now though, having stepped back 'out of the weeds' - I probably shouldn't design for a non-deterministic way of using the tracker. I should probably both document and test things to use the preferable approach, and leave those who don't follow it to their own devices...

index.js
colmsnowplow: Haha damn I thought I had removed this comment! Thanks!

README.md
colmsnowplow: I think I copied how the iOS tracker refers to it. But this makes sense - the android tracker I believe is just `platformContext`, and the convention you describe is one I'm on board with!

ios/RNSnowplowTracker.m
colmsnowplow: After some testing, I was actually wrong - on configuring it with multiple error-invoking values, it attempts to reject multiple times, and crashes. I'll change it to return on error.

colmsnowplow:
OK @istreeter @AlexBenny , I believe I've addressed everything - I tried to keep things auditable but unfortunately did have to force push to fix a conflict. Apologies for that.

istreeter (approved):

README.md
```diff
@@ -103,41 +111,88 @@ setSubjectData({ // All parameters optional
...
 ```
 
 ### Track a Page View:
 
 ```JavaScript
-trackPageViewEvent('my-page-url.com', // required, page url, string
-                  'my page title', // optional - set to null if not needed, page title, string
-                  'my-page-referrer.com', // optional - set to null if not needed, referrer url, string
-                  []); // array of self-describing JSONs for custom contexts, or an empty array if none are to be attached
+Tracker.trackPageViewEvent({
```
paulboocock: This example is different to all the rest. It's `Tracker.track...` whereas all the other examples are `track...`

paulboocock (approved):

colmsnowplow:
@elliotstokes finally ready for your final pass over this PR. Apologies for delay, rebasing was more complicated than I had anticipated.

In your own time, would v much appreciate it if you gave this a once over. :)

index.js
```diff
@@ -1,6 +1,73 @@
...
+                 lifecycleEvents : false,
+                 screenContext : false,
+                 sessionContext : false,
+                 foregroundTimeout : 600,
+                 backgroundTimeout : 300,
+                 checkInterval : 15,
+                 installTracking : false}
+
+    if (typeof argmap.endpoint !== 'undefined' && typeof argmap.appId !== 'undefined' && typeof argmap.namespace !== 'undefined' ) {
+      return RNSnowplowTracker.initialize({...defaults, ...argmap});
```
elliotstokes: Potentially a dev experience improvement here would be to manage or hold an instance of this `initPromise` internally as well so when `trackScreenViewEvent` was called you could await it in that function internally rather than the scenario where a user forgets to await it outside and calls a track method before the promise resolves and it causes issues.

elliotstokes (approved):
Looks good. Just that one suggestion about a potential improvement.

index.js
colmsnowplow: Ty Elliot. Shame not to have time to do this in this PR, but I've logged it in [an issue](https://github.com/snowplow-incubator/snowplow-react-native-tracker/issues/78) for next release!

MarlosAugusto:
Hello @colmsnowplow , what is the status of these implementations? I am working on a project that intends to use snowplow on mobile, but so far it is not 100% (we already use it in WEB projects), we are waiting to test its use in the new version.

PR #107, opened 2021-03-19
Release/0.1.4

colmsnowplow:
Just a quick bugfix release to update demo App and address this issue: https://github.com/snowplow-incubator/snowplow-react-native-tracker/issues/105


package.json
```diff
@@ -15,8 +15,8 @@
   ],
   "license": "Apache-2.0",
   "peerDependencies": {
-    "react-native": "^0.59.8",
-    "react": "^16.8.6"
+    "react-native": ">=0.59.8",
```
AlexBenny, line 18: I really dislike the way the manage versioning. Technically it's impossible to know whether there are breaking changes 🤦.

AlexBenny (approved):
Looks good! 👍 🚢

package.json
colmsnowplow: Yeah I don't really like it either, I think this is why I fixed to major version before.

colmsnowplow:
Made a mistake in publishing and so had to change version to 0.1.5. Closing this in favour of https://github.com/snowplow-incubator/snowplow-react-native-tracker/pull/108 because of that.

PR #123, opened 2021-05-21
Release/0.1.7

adatzer:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,136 @@
...
+    - name: Setup Node
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14
+
+    - name: Npm ci
+      run: npm ci
+
+    - name: NPM Publish
+      uses: JS-DevTools/npm-publish@v1
```
paulboocock: Any reason why we're using `JS-DevTools/npm-publish@v1` rather than a straight up `npm publish`? I have no issue with this, it looks pretty well maintained, I was just curious what benefits it added?

paulboocock (approved):
LGTM

.github/workflows/deploy.yml
adatzer: I just liked the features it includes, like the minor version check with npm's latest and the security one (not writing the secret). I am also a bit worried for trying it, so i can certainly replace it with an `npm publish`.

paulboocock: No reason to switch it out, it looks like a useful tool. Happy to give it a try and see how it goes, likely allow us to move a little faster than using `npm publish` directly.

PR #127, opened 2021-06-16
Release/0.2.0

adatzer:


DemoApp/App.js
```diff
@@ -19,15 +19,9 @@ import {
...
-  Colors,
-  DebugInstructions,
-  Header,
-  LearnMoreLinks,
-  ReloadInstructions,
-} from 'react-native/Libraries/NewAppScreen';
+import {Colors, Header} from 'react-native/Libraries/NewAppScreen';
 
-import Tracker from '@snowplow/react-native-tracker';
+import * as Snowplow from '@snowplow/react-native-tracker';
```
paulboocock: As an example, I think it would be nicer to:

```suggestion
import { createTracker } from '@snowplow/react-native-tracker';
```

README.md
```diff
@@ -20,34 +20,42 @@ From the root of your react-native project:
...
-  lifecycleEvents : false,
-  screenContext : true,
-  sessionContext : true,
-  foregroundTimeout : 600,
-  backgroundTimeout : 300,
-  checkInterval : 15,
-  installTracking : false
-});
+```JavaScript
+import * as Snowplow from '@snowplow/react-native-tracker';
```
paulboocock: Same comment as above, import just the function rather than the whole library.

README.md
```diff
@@ -20,34 +20,42 @@ From the root of your react-native project:
...
+
+const tracker = Snowplow.createTracker('my-namespace', {
```
paulboocock: Which means this line becomes:
```suggestion
const tracker = createTracker('my-namespace', {
```

README.md
```diff
@@ -20,34 +20,42 @@ From the root of your react-native project:
...
+    sessionContext: true,
+    foregroundTimeout: 600,
+    backgroundTimeout: 300,
+    checkInterval: 15,
+    installTracking: false,
+  });
+
+tracker.trackSelfDescribingEvent(
+  {
+    'schema': 'iglu:com.acme/hello_world_event/jsonschema/1-0-0',
```
paulboocock: Why the quotes on `'schema'`? It's not wrong but seems inconsistent with the object above. I don't usually add quotes to the keys of an object.

README.md
```diff
@@ -166,70 +148,113 @@ trackPageViewEvent({
...
 
 ```JavaScript
-trackSelfDescribingEvent({ // Self-describing JSON for the custom event
-  'schema': 'iglu:com.acme/event/jsonschema/1-0-0',
-  'data': {'message': 'hello world'}
+
+trackSelfDescribingEvent(
+  // Self-describing JSON for the custom event
+  {
+    'schema': 'iglu:com.acme/event/jsonschema/1-0-0',
```
paulboocock: nit: same here on the quotes unlike the other examples below.

README.md
```diff
@@ -166,70 +148,113 @@ trackPageViewEvent({
...
 
 ## Contributing
 
 Please read CONTRIBUTING.md for general guidelines on contributing.
 
-### Setting up
+### Maintainer quick start
 
-Assuming a react-native environment is set up, the demoApp can be used to test changes, with:
+Assuming a react-native environment is set up, from the root of the repository to launch the DemoApp:
```
paulboocock: Might be worth adding a link to reacts documentation on setting up on the `react-native environment` text?

package.json
```diff
@@ -19,9 +23,27 @@
...
+  },
   "peerDependencies": {
-    "react-native": ">=0.59.8",
-    "react": ">=16.8.6"
+    "react": ">=16.8.6",
+    "react-native": ">=0.59.8"
   },
-  "dependencies": {}
+  "devDependencies": {
+    "@babel/core": "^7.14.5",
```
paulboocock: Why do we include `babel` if we're using tsc to compile (`"compile": "rm -rf dist && tsc -p .",`)?
I'd expect to use tsc with no emit, just for type checking, and then use babel for compilation if we need to use babel. Although I'd check if we 100% even need babel or if we can use `tsc` by itself.

src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+  return Object.freeze({
+    setSubjectData,
+    trackScreenViewEvent,
+    trackSelfDescribingEvent,
+    trackStructuredEvent,
+    trackPageViewEvent,
+  });
+}
+
+export { createTracker };
```
paulboocock, line 59: We should also export the interfaces and types (like `StructuredProps`) so users can use them to make their own objects when calling functions like `trackStructuredEvent`

src/types.ts
```diff
@@ -0,0 +1,325 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+
+/**
+ * HttpMethod type
+ */
+type HttpMethod = 'post' | 'get';
```
paulboocock: It's worth exporting these too. You never know who will want it 🤷

tsconfig.json
```diff
@@ -0,0 +1,52 @@
...
+    "allowJs": true,
+    "jsx": "react-native",
+    "types": ["react", "react-native"],
+    "declaration": true,
+    "sourceMap": true,
+    "isolatedModules": true,
+    "noEmitOnError": true,
+
+    /* Strict Type-Checking Options */
+    "strict": true,
```
paulboocock, line 12: Because `strict` is true, you can remove some of the options below as they will become true by default. Looking at https://www.typescriptlang.org/tsconfig#strict and searching for `unless strict is set` should give you a good idea. Not essential though and some people do prefer being more explicit around these options, your call!

paulboocock (commented):
Looking good! Couple of comments, some we've already discussed. No major blockers here once those couple of things I've mentioned have been considered.

package.json
adatzer: Thank you for reviewing @paulboocock !
Indeed, i had blindly left `babel` around. Now removed it completely as we use `tsc` for compilation. Also left the declarations to [dts-bundle-generator](https://github.com/timocov/dts-bundle-generator) (thanks for the pointer!) so that they now get all bundled into `index.d.ts`.
For ref, the last force-push also removed the redundant strict options from `tsconfig.json`, exported the interfaces and types, added links to Readme, removed quoting object keys and switched to `import {createTracker} ..` both in readme examples and DemoApp.

package.json
```diff
@@ -1,11 +1,23 @@
...
   "name": "@snowplow/react-native-tracker",
-  "version": "0.1.7",
+  "version": "0.2.0",
   "description": "A library for tracking Snowplow events in React Native",
-  "main": "index.js",
+  "homepage": "https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/react-native-tracker/",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
+    "compile": "rm -rf dist && tsc -p .",
```
paulboocock: I'd suggest switching to [rimraf](https://www.npmjs.com/package/rimraf) CLI as a dev dependency to remove these files and folders in the `compile` and `build` commands as `rm -rf` is a unix command so won't work on Windows.

paulboocock (approved):
LGTM. One last optional comment, feel free to open as an issue for v1 rather than here if you'd like to release now, your call.

PR #128, opened 2021-07-22
Release/1.0.0

adatzer:
Starting as a draft PR, because of the things still left to do:
1. change e2e tests after upgrade to v2
2. rest of issues of milestone, some of which seem resolved by the upgrade (to be confirmed)


.eslintrc.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     },
     sourceType: 'module',
   },
-  ignorePatterns: ['.eslintrc.js', 'rollup.config.js'],
+  ignorePatterns: ['.eslintrc.js', 'rollup.config.js', 'tests/**/*'],
```
paulboocock: Any reason we're not linting tests? Seems reasonable to lint the tests to me but open to why we might not want to.

.github/workflows/build.yml
```diff
@@ -27,11 +27,14 @@ jobs:
 
     - name: Install node modules for DemoApp
       working-directory: DemoApp
-      run: yarn install --frozen-lockfile
+      run: yarn install --frozen-lockfile --network-timeout 100000
```
paulboocock: Why do we install the demo app dependencies to test the tracker? Seems like an unnecessary step to me.

.github/workflows/e2e-android.yml
```diff
@@ -0,0 +1,142 @@
...
+    - name: Setup Node
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14
+
+    - name: Setup Java
+      uses: actions/setup-java@v2
+      with:
+        distribution: 'adopt'
+        java-version: '8'
```
paulboocock, line 25: Does this work with 11 instead? 8 is pretty old now.

.github/workflows/e2e-android.yml
```diff
@@ -0,0 +1,142 @@
...
+
+    # -- Micro --
+    - name: Cache Micro
```
paulboocock, line 28: This is clever!

.github/workflows/e2e-android.yml
```diff
@@ -0,0 +1,142 @@
...
+          yarn test
+
+    - name: Log events if Detox-Jest failed
+      if: ${{ failure() && steps.dj.outcome == 'failure' }}
+      run: |
+        echo '-- Bad events --'
+        curl -s -o - 'http://0.0.0.0:9090/micro/bad' | jq '.'
+        echo '-- Good events --'
+        curl -s -o - 'http://0.0.0.0:9090/micro/good' | jq '.'
+
```
paulboocock: I love this idea! The output could be huge though, I wonder if piping it to a file and uploading it as a build artifact would be easier as you could pull it into an IDE then.

DemoApp/android/build.gradle
```diff
@@ -9,6 +9,7 @@ buildscript {
         ndkVersion = "20.1.5948944"
     }
     repositories {
+        mavenCentral()
         google()
         jcenter()
```
paulboocock, line 14: nit: You should be able to remove jcenter() now.

DemoApp/android/build.gradle
```diff
@@ -30,9 +31,12 @@ allprojects {
             // Android JSC is installed from npm
             url("$rootDir/../node_modules/jsc-android/dist")
         }
-
+        mavenCentral()
         google()
         jcenter()
```
paulboocock, line 36: nit: same as above

DemoApp/tests/e2e/helpers/microHelpers.js
```diff
@@ -0,0 +1,319 @@
...
+  } else if (expType === '[object Object]') {
+    return Object.keys(expVal).every(function (k) {
+      return Object.keys(actVal).includes(k) && compare(expVal[k], actVal[k]);
+    });
+  } else {
+    return expVal === actVal;
+  }
+}
+
+export {
```
paulboocock, line 309: This is a nice file! Nice one.

android/src/main/java/com/snowplowanalytics/react/util/ConfigUtil.java
```diff
@@ -0,0 +1,343 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.HashMap;
+import java.util.List;
+import java.util.ArrayList;
+
+import com.snowplowanalytics.react.util.EventUtil;
+
+public class ConfigUtil {
+
+    private final static String RN_TRACKER_VERSION = "rn-0.2.0";
```
paulboocock: This is a well hidden, easy to miss, string. Possible to move it to a Constants file or a Version file?

android/src/main/java/com/snowplowanalytics/react/util/EventUtil.java
```diff
@@ -2,115 +2,200 @@
...
-            SelfDescribingJson json = EventUtil.getSelfDescribingJson(contexts.getMap(i));
+            SelfDescribingJson json = getSelfDescribingJson(contexts.getMap(i));
             nativeContexts.add(json);
         }
+
         return nativeContexts;
     }
 
-    public static SelfDescribingJson getSelfDescribingJson(ReadableMap json) {
+    public static Structured getStructuredEvent(ReadableMap argmap) {
```
paulboocock: nit: Should the methods in this file be `create...` or `build...` rather than `get...`? Feels like we'd constructing new objects rather than retrieving existing ones.

android/src/main/java/com/snowplowanalytics/react/util/EventUtil.java
```diff
@@ -2,115 +2,200 @@
...
+            eventBuilder.documentName(argmap.getString("name"));
+        }
+        if (argmap.hasKey("documentDescription")) {
+            eventBuilder.documentDescription(argmap.getString("documentDescription"));
+        }
+
+        return eventBuilder.build();
+    }
+
+    public static List<EcommerceTransactionItem> getEcomItems(ReadableArray items) {
```
paulboocock:
```suggestion
    public static List<EcommerceTransactionItem> getEcommerceTransactionItems(ReadableArray items) {
```

src/constants.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+
+'use strict';
+
+const spLog = {
```
paulboocock: nit: `spLogMessages` might be a better name? `spLog` sounds like the logger itself. You could probably drop the `sp` too, `logMessages` works for me.

src/events.ts
```diff
@@ -0,0 +1,272 @@
...
+  validateSelfDesc,
+  validateScreenView,
+  validateStructured,
+  validatePageView,
+  validateTiming,
+  validateConsentGranted,
+  validateConsentWithdrawn,
+  isValidEcomItem,
+  validItemsArg,
+  validateEcommerceTransaction,
```
paulboocock, line 271: I think this is probably the only tracker that prevalidates events before sending them. I see some value in it but I'm still not 100% convinced. My main concern is that if we change an event definition (especially those that are SelfDescribing under the hood) then we need to remember to change this or else it won't work. Some of these calls aren't cheap either, iterating through the entire list of ecom items is a good example. The big question is should be do this in the tracker to catch (some) bad events earlier or should we defer this always to the pipeline to validate events? 
I remember us discussing this in a call, but I think it would be worth recapping the thoughts on why we might want to validate events in this tracker before passing the events to the iOS and Android trackers, whereas we don't do this in the "native" trackers.

src/subject.ts
```diff
@@ -0,0 +1,265 @@
...
+  networkUserId: setNetworkUserId,
+  domainUserId: setDomainUserId,
+  ipAddress: setIpAddress,
+  useragent: setUseragent,
+  timezone: setTimezone,
+  language: setLanguage,
+  screenResolution: setScreenResolution,
+  screenViewPort: setScreenViewPort,
+  colorDepth: setColorDepth
+} as Record<keyof SubjectConfiguration, <Type>(ns: string, x: Type | null) => Promise<void>>;
```
paulboocock: Sometimes TypeScript is both wonderful and awful at the same time

src/subject.ts
```diff
@@ -0,0 +1,265 @@
...
+/**
+ * Sets the tracker subject
+ *
+ * @param ns {string} - the tracker namespace
+ * @param subjData {SubjectConfiguration} - the new subject data
+ * @returns - Promise
+ */
+function setSubjectData(
+  ns: string,
+  subjData: SubjectConfiguration
```
paulboocock:
```suggestion
  subject: SubjectConfiguration
```

paulboocock: or `config`

src/tracker.ts
```diff
@@ -0,0 +1,251 @@
...
+ *
+ * @param ns {string} - the tracker namespace
+ * @param argmap {Object} - the event data
+ * @param ctxt {Array}- the event contexts
+ * @returns {Promise}
+ */
+function trackSelfDescribingEvent(
+  ns: string,
+  argmap: SelfDescribing,
+  ctxt: EventContext[] = []
```
paulboocock: Where did all the vowels go? 
I think I'd extend `ns` to `namespace` too. No need to be short here, longer names are always more descriptive and easier to work with.

src/types.ts
```diff
@@ -17,63 +17,301 @@
...
+
+/**
+ * BufferOption
+ */
+export type BufferOption = 'single' | 'default' | 'large';
+
+/**
+ * Size
+ */
+export type Size = [number, number];
```
paulboocock: I'm commenting here since I'm reading through. I could figure out what this used for but `Size` is rather generic, can we think of a more descriptive name?

src/types.ts
```diff
@@ -149,33 +389,163 @@ export type PageViewProps = {
...
+   * Tracks a self-descibing event
    *
-   * @param argmap - The subject data to be set or updated
+   * @param argmap - The self-describing event properties
+   * @param ctxt - The array of event contexts
    */
-  readonly setSubjectData: (argmap: SubjectData) => Promise<void>;
+  readonly trackSelfDescribingEvent: (
+    argmap: SelfDescribing,
+    ctxt: EventContext[]
```
paulboocock: nit: More missing vowels ;-)

src/types.ts
```diff
@@ -220,106 +579,181 @@ export type ReactNativeTracker = {
...
-export interface InitTrackerConfig extends TrackerConfig {
   /**
-   * The tracker namespace
+   * Sets the screenViewPort of the tracker subject
+   *
+   * @param newView - The new screenViewPort
    */
-  namespace: string;
-}
+  readonly setScreenViewPort: (newView: Size | null) => Promise<void>;
```
paulboocock: nit: Viewport is one word

src/types.ts
```diff
@@ -220,106 +579,181 @@ export type ReactNativeTracker = {
...
 
   /**
-   * The app ID
+   * Sets the screenResolution of the tracker subject
+   *
+   * @param newRes - The new screenResolution
    */
-  appId: string;
-}
+  readonly setScreenResolution: (newRes: Size | null) => Promise<void>;
```
paulboocock: Seeing `Size` here, I wonder if `Resolution` or `ScreenSize` is a better name.

DemoApp/ios/Podfile.lock
```diff
@@ -319,11 +319,11 @@ PODS:
     - React-cxxreact (= 0.64.1)
     - React-jsi (= 0.64.1)
     - React-perflogger (= 0.64.1)
-  - RNSnowplowTracker (1.0.0):
+  - RNSnowplowTracker (0.2.0):
```
paulboocock: Will we bump this to 1.0.0 for release?

paulboocock (commented):
This is an amazing PR, lots of great things in here from E2E tests to the updated API. I've made a few minor comments but this is looking in great shape. 
I'm looking forward the final PR, looks like there isn't much left to do now bar figuring out that iOS e2e test failure.

.eslintrc.js
adatzer: Thank you for reviewing @paulboocock !
There is indeed no reason for not linting the tests. Adding now!

.github/workflows/e2e-android.yml
adatzer: This is because of the Detox android setup guide that explicitly mentions it as required [here](https://github.com/wix/Detox/blob/master/docs/Introduction.AndroidDevEnv.md#java-setup)

.github/workflows/e2e-android.yml
adatzer: Indeed, as an artifact it would be much more easy to inspect the events in case of failure!

DemoApp/ios/Podfile.lock
adatzer: Yes, i was thinking to leave the change for the prepare-for-release commit, so that all the version changes appear together. The version would need to be changed in:
 - package.json
 - the relevant android-bridge file
 - the relevant ios-bridge file
 - Podfile.lock in DemoApp/ios

paulboocock: Sounds good to me!

src/events.ts
adatzer: Indeed, i've been changing my mind on this. 
On one hand the checks are redundant, since there is some validation in the native trackers too. 
On the other hand, some parts of this validation offer some guarantees in the bridge, so that we don't have to, for example, check if arrays are arrays or have if-else statements even for the required event props in order to construct the event. 

For example, a screenViewEvent without `name` at the moment will not reach the bridge at all. But if we removed the check in JavaScript-side, then it would need to be done in the bridge-side (i'll have another look to confirm my assumption here). Would that be better from performance perspective? 
Or do you mean to just let it error in the bridge? That would be fine for the Android bridge, but i was not confident of error handling in ObjC, where try-catching somehow seemed to me hard to trust.

Then as you said, there is the bad events side of things. At the moment, there are only checks for the required props. And that is only for the javascript users that have no type-checking, or for a runtime edge case. Not much of an argument for checking indeed here.

Concerning possible changes to event definitions, would that be a breaking change in the native trackers?
Generally, is there a right balance? What do you think?

.github/workflows/build.yml
adatzer: I had forgotten why that was "needed". In fact it is only for linting to go through to the DemoApp directory, since it is not ignored  from root eslint. It is indeed an unnecessary step there, as we shouldn't eitherway need DemoApp deps to lint tracker code.

DemoApp/android/build.gradle
adatzer: It turned out this is not possible just yet, because of some react-native and flipper dependencies. This is also being discussed in:
https://github.com/facebook/yoga/issues/1065
https://github.com/react-native-community/discussions-and-proposals/issues/331

paulboocock: Ok. I think it's worth opening a issue in this repo to remind us to remove this when we can. It'll likely stop building once jcenter is fully offline.

rollup.config.js
```diff
@@ -1,12 +1,15 @@
...
+import sourcemaps from 'rollup-plugin-sourcemaps';
 const config = [
   {
     input: './dist/tmp/index.js',
     output: {
       file: 'dist/index.js',
       format: 'es',
+      sourcemap: true,
     },
+    plugins: [sourcemaps()],
```
paulboocock: You're going to want to mark `react-native` as an external dependency since its a peerDependency (and it'll also remove the rollup warning).

```suggestion
    input: './dist/tmp/index.js',
    output: {
      file: 'dist/index.js',
      format: 'es',
      sourcemap: true,
    },
    external: [ 'react-native' ],
    plugins: [sourcemaps()],
```

package.json
```diff
@@ -10,13 +10,14 @@
...
   },
   "files": [
-    "src",
     "dist",
     "android",
-    "ios"
+    "ios",
+    "RNSnowplowTracker.podspec"
   ],
   "keywords": [
```
paulboocock, line 22: Feel free to add yourself to the `contributors` list!

src/events.ts
paulboocock: I'm happy for this to stay in, it isn't too heavy and seems safer to have it than not given the brigding in this application. Even if a schema did update, you'd need to update the native tracker and therefore update this tracker too. Perhaps something we can look in to more in the future but these feels like a good compromise at the moment.

paulboocock (approved):
LGTM. Great PR, love all the new testing options here.

The README needs updating, I think you're already planning to remove the code from it and move it to the docs which I think is a good idea. The README is also missing a copyright notice, you can steal that from something like: https://github.com/snowplow/snowplow-objc-tracker/
Also, as this has deviated quite a lot from the original fork, I think we should update the copyright notice. We should still include DataCamp's copyright, but we can flip it to be more like this: https://github.com/snowplow/snowplow-javascript-tracker/blob/master/LICENSE#L3-L4
We can do that in the README and in LICENSE (at the bottom of the Apache 2.0 file).

paulboocock (approved):
LGTM! Assuming the `beta.1` Github release works this time ;-) 

Good updates to README.md, looks great.

PR #139, opened 2022-01-19
Update demo app to show accessing tracker instance from platform native code (close #136)

### LINKED ISSUE #136: Update demo app to show accessing tracker instance from platform native code
For apps that contain both native and react native code and want to be able to track events from both parts using the React tracker instance, we should come up with an approach to achieve this. It might be already possible to do out of the box, but we need to identify what would be the best approach.

Extend the demo app to showcase this.

matus-tomlein:
## Track events from native code in demo app

This PR adds an example usage of the native tracker APIs (for Android and iOS) in the demo app. It listens for key presses in the `MainActivity.java` under Android and `ViewController.m` under iOS and tracks simple structured events on each key press. This shows that trackers instantiated in React Native can also be accessed in platform native code using the mobile tracker APIs.

I also a added a section in the documentation that discusses the issue and refers to this demo app example. There is a PR for it in the data-value-resources repository: https://github.com/snowplow-incubator/data-value-resources/pull/37

## Compilation errors under ARM Macs

Also made a small change in the XCode project for the iOS code of the tracker because of some compilation issues under my ARM Mac.

### DIFF
--- a/DemoApp/android/app/src/main/java/com/demoapp/MainActivity.java
+++ b/DemoApp/android/app/src/main/java/com/demoapp/MainActivity.java
@@ -1,6 +1,10 @@
 package com.demoapp;
 
+import android.view.KeyEvent;
+
 import com.facebook.react.ReactActivity;
+import com.snowplowanalytics.snowplow.Snowplow;
+import com.snowplowanalytics.snowplow.event.Structured;
 
 public class MainActivity extends ReactActivity {
 
@@ -11,5 +15,14 @@ public class MainActivity extends ReactActivity {
   @Override
   protected String getMainComponentName() {
     return "DemoApp";
+  }
+
+  /**
+   * Demonstrates the use of a tracker initialized in React native.
+   */
+  @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
+    Snowplow.getDefaultTracker().track(new Structured("key", "press"));
+    return super.onKeyDown(keyCode, event);
   }
 }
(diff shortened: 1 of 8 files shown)


DemoApp/android/app/src/main/java/com/demoapp/MainActivity.java
```diff
@@ -12,4 +16,13 @@
   protected String getMainComponentName() {
     return "DemoApp";
   }
+
+  /**
+   * Demonstrates the use of a tracker initialized in React native.
+   */
+  @Override
+  public boolean onKeyDown(int keyCode, KeyEvent event) {
+    Snowplow.getDefaultTracker().track(new Structured("key", "press"));
```
adatzer, line 25: Just to confirm: Someone could also use `getTracker` and get the tracker by namespace as well, right?

adatzer (approved):
LGTM! The example is really great!

Just on the nits of it:
1. this branch will get eventually merged on a release one, right?
2. At the moment the arm-commit contains example code, so just to confirm, is a different issue for arm Macs compilation error to be made or do you plan to squash the commits? 
3. do you think it makes sense to specify in the issue title that the native access is enabled in the demo app for clarity?

DemoApp/android/app/src/main/java/com/demoapp/MainActivity.java
matus-tomlein: Yes, that should work as well.

matus-tomlein:
Thanks for the review @adatzer! I made some updates based on your suggestions:

1. I changed the base branch to release 1.1.0.
2. I created an issue for the compilation error fix and referenced it in this commit. I plan to rebase the commits from this branch to the release one so that both commits will remain independent.
3. Also changed the issue title to make it more clear that it is an update to the demo app.

PR #144, opened 2022-01-23
Update mobile tracker dependencies to v3 (close #141)

### LINKED ISSUE #141: Update mobile tracker dependencies to v3

matus-tomlein:
This PR updates the dependencies for native mobile trackers to version 3. It also removes use of deprecated APIs that were removed.

Issue #141.

### DIFF
--- a/android/src/main/java/com/snowplowanalytics/react/util/EventUtil.java
+++ b/android/src/main/java/com/snowplowanalytics/react/util/EventUtil.java
@@ -4,6 +4,10 @@ import androidx.annotation.NonNull;
 
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.ReadableArray;
+import com.snowplowanalytics.snowplow.event.DeepLinkReceived;
+import com.snowplowanalytics.snowplow.event.MessageNotification;
+import com.snowplowanalytics.snowplow.event.MessageNotificationAttachment;
+import com.snowplowanalytics.snowplow.event.MessageNotificationTrigger;
 import com.snowplowanalytics.snowplow.event.Structured;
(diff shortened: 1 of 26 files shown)


android/src/main/java/com/snowplowanalytics/react/util/EventUtil.java
```diff
@@ -37,165 +39,174 @@ public static SelfDescribingJson createSelfDescribingJson(ReadableMap json) {
     }
 
     public static Structured createStructuredEvent(ReadableMap argmap) {
-        Structured.Builder eventBuilder = Structured.builder()
-            .category(argmap.getString("category"))
-            .action(argmap.getString("action"));
+        Structured event = new Structured(
+                Objects.requireNonNull(argmap.getString("category"), "category can't be null"),
```
adatzer, line 47: Just this question (applies for all the `requireNonNull`'s added): why we add this extra check, since we already check/validate in RN side, for example using this [here](https://github.com/snowplow/snowplow-react-native-tracker/blob/809d74f50607f751283ee5b60c4f9b8d501ee5ca/src/events.ts#L101)?

adatzer (commented):
Looks great!

android/src/main/java/com/snowplowanalytics/react/util/EventUtil.java
matus-tomlein: Yeah, you're right, it's not really necessary. I was just trying to resolve the warnings that I was getting in Android Studio because they were annoying me. I guess it doesn't hurt to double check although it is better that we check sooner on the RN side.

adatzer (approved):

PR #145, opened 2022-01-23
Fix check for byteLimitGet when reading emitter configuration (close #138)

### LINKED ISSUE #138: Fix check for byteLimitGet when reading emitter configuration
The if condition checks for "byteLimitPost" instead of "byteLimitGet": https://github.com/snowplow/snowplow-react-native-tracker/blob/master/android/src/main/java/com/snowplowanalytics/react/util/ConfigUtil.java#L196

matus-tomlein:
This is a small change in reading the emitter configuration in Java where a wrong attribute was read.


adatzer (approved):
LGTM!

Issue #142, opened 2022-01-21
Add deep link and message notification events


matus-tomlein:
This was introduced in v3 of mobile trackers.

PR #147, opened 2022-01-23
Add deep link and message notification events (close #143, #142)

### LINKED ISSUE #143: Add deep link event and context
This was introduced in v3 of mobile trackers.

matus-tomlein:
This PR adds new event types introduced in the v3 of native trackers:

* deep link event and context – can be used by users to track when deep links are received. The context is attached to subsequent screen views tracked after the deep link received event,
* message notification event to track received notifications by users.

### DIFF
--- a/ios/RNSnowplowTracker.m
+++ b/ios/RNSnowplowTracker.m
@@ -44,6 +44,7 @@
 #import <SnowplowTracker/SPStructured.h>
 #import <SnowplowTracker/SPEcommerceItem.h>
 #import <SnowplowTracker/SPEcommerce.h>
+#import <SnowplowTracker/SPDeepLinkReceived.h>
 
 
 @implementation RNSnowplowTracker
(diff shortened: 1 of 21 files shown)


tests/__tests__/events.test.ts
```diff
@@ -394,6 +394,34 @@ describe('test validateConsentWithdrawn', () => {
...
+  });
+
+  test('test invalid deepLinkReceived - invalid url', async () => {
+    const testEv = {
+      url: true
+    } as any;
+    await expect(e.validateDeepLinkReceived(testEv)).rejects.toThrow(logMessages.deepLinkReq);
+  });
+
+  test('test valid consentGranted', async () => {
```
adatzer: nit - `'test valid deepLinkReceived'`

src/types.ts
```diff
@@ -532,6 +537,113 @@ export type EcommerceTransactionProps = {
...
+   */
+   titleLocArgs?: string[];
+  /**
+   * The key to the title string in the app's string resources to use to localize the title text to the user's current localization.
+   */
+   titleLocKey?: string;
+  /**
+   * The trigger that raised the notification message. Must be one of: push, location, calendar, timeInterval, other
+   */
+   trigger: string;
```
adatzer: Would it make sense to make a `Trigger` type like we have for other enums?

src/events.ts
```diff
@@ -202,6 +204,51 @@ function validateConsentWithdrawn(argmap: ConsentWithdrawnProps): Promise<boolea
...
+  // validate type
+  if (!isObject(argmap)) {
+    return Promise.reject(new Error(logMessages.evType));
+  }
+  // validate required props
+  if (
+    typeof argmap.title !== 'string'
+    || typeof argmap.body !== 'string'
+    || typeof argmap.trigger !== 'string'
+    || !['push', 'location', 'calendar', 'timeInterval'].includes(argmap.trigger)
```
adatzer: `'other'` is also one of the enum values for trigger

ios/RNSnowplowTracker.m
```diff
@@ -433,6 +435,148 @@ @implementation RNSnowplowTracker
...
+        }
+        NSString *bodyLocKey = [argmap sp_stringForKey:@"bodyLocKey" defaultValue:nil];
+        if (bodyLocKey) {
+            event.bodyLocKey = bodyLocKey;
+        }
+        NSString *category = [argmap sp_stringForKey:@"category" defaultValue:nil];
+        if (category) {
+            event.category = category;
+        }
+        NSNumber *contentAvailable = [argmap sp_numberForKey:@"contentAvailable" defaultValue:nil];
```
adatzer, line 526: not sure how this plays out with booleans, how about `sp_boolForKey`?

adatzer (commented):
Looks great @matus-tomlein !

tests/__tests__/events.test.ts
matus-tomlein: Thanks, good catch!

src/types.ts
matus-tomlein: I was considering enums but didn't notice how they are implemented in types.ts, that's much better! Changed it, thanks.

src/events.ts
matus-tomlein: Lifesaver!

ios/RNSnowplowTracker.m
matus-tomlein: NSNumber also supports booleans. We could also use `sp_boolForKey` but the `event.contentAvailable` property takes NSNumber so we would have to transform to NSNumber anyway.

adatzer (approved):
LGTM!!

PR #155, opened 2022-03-03
Add support for React Native for tvOS (close #154)

### LINKED ISSUE #154: Add support for React Native for tvOS
Investigate steps needed to support the [React Native TVOS](https://github.com/react-native-tvos/react-native-tvos) fork of React Native.

This [Discourse post](https://discourse.snowplowanalytics.com/t/react-native-snowplow-tvos/6444) indicates issues with initialization of the tracker under TVOS.

matus-tomlein:
This PR adds support for tvOS and Android TV using the react-native-tvos fork of React Native (issue #154). The only real change that had to be done was adding "tvos" to the supported platforms in the podspec for iOS and adding a Demo app and all the credit goes to @adatzer. It also works on Android TV without any changes (one only has to install the emulator for Android TV).

I was trying to make the end-to-end tests work and almost got it to work but couldn't make Detox send any actions to the simulator. Detox launched and the app was ready but there is no way to control the TV interface using Detox. I tried making keyboard actions to trigger the buttons but they didn't get through. Detox currently [doesn't support tvOS and doesn't plan to add support](https://github.com/wix/Detox/discussions/2853) so I don't see a way to make the e2e test work here.

Not sure about what release version to go for. Since this is a new functionality, I thought it should be a minor release – 1.2.0. But we can also make it a patch release (1.1.1).


README.md
```diff
@@ -105,6 +105,30 @@ bash .scripts/cleanBuildAndRun.sh ios
...
+```bash
+yarn android
+```
+
+_Note_: Linux users who want to run the DemoApp for Android, would also need to run `yarn start` in a separate terminal.
+
+**For tvOS:**
+
+```bash
+yarn pods && yarn tvos
```
AlexBenny, line 134: I don't think `yarn pods` work. Am I missing something?
I needed to launch `pod update` from the `ios` subfolder.

Also `yarn tvos` launches `xcodebuild` with the parameters about the simulator `Apple TV`. In my case the name is different `Apple TV (15.2)` probably because I installed the simulator later. I don't know if there is a better way to select the simulator. However, I'm not sure that's the problem because I've changed the name in the `package.json` and relaunched `yarn tvos` and then it complained that `'DemoAppTV.xcworkspace' does not exist.`. This because we need to switch to the subfolder where it can find the workspace.

Let me know if I just missed a step somewhere and I'm in the wrong track!

AlexBenny (commented):
Probably it's something wrong with my setup but I'm struggling to launch the demo app for tvOS.

README.md
matus-tomlein: Sorry, I must have missed adding the `pods` script to the package.json for the demo app. I have just added it there, can you please try again with the `yarn pods` command and `yarn tvos` after it? I think that should help with the missing `DemoAppTV.xcworkspace`.

Not sure how to solve the different simulator name – if I don't put the simulator name there, it will launch an iPhone simulator by default, so I have to put there something. I added a note to the README to update the package.json in case of different simulator.

Also added a note to README to update the collector endpoint in `DemoAppTV/App.js`.

DemoAppTV/App.js
```diff
@@ -0,0 +1,486 @@
...
+  const tracker = createTracker(
+    'sp1',
+    {
+      endpoint: 'placeholder',
+    },
+    {
+      trackerConfig: {
+        appId: 'DemoAppId',
+        base64Encoding: false,
+        devicePlatform: 'iot',
```
adatzer: What do you think about setting as `devicePlatform: 'tv'`?

adatzer (approved):
Looks great!!

DemoAppTV/App.js
matus-tomlein: Thanks Ada, will fix this!

PR #157, opened 2022-03-03
Bump min RN version supported to v0.65+ (close #153)

### LINKED ISSUE #153: Bump min RN version supported to v0.65+
As reported here (#135) the Android tracker could crash due to conflict with the OkHttp library version as React Native internally uses OkHttp. The React Native tracker v1+ requires the Android Tracker v2+ which requires OkHttp v4+.
React Native supports OkHttp v4+ from the version 0.65.

We need to update the tracker and related documentation indicating we support React Native v0.65+.

matus-tomlein:
* Bump min react-native version to 0.65 according to issue #157 
* Update versions for all dependencies to latest versions

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,8 +44,8 @@
   },
   "devDependencies": {
     "@types/jest": "^27.4.1",
-    "@types/react": "^17.0.39",
-    "@types/react-native": "^0.67.1",
+    "@types/react": "^16.8.25",
+    "@types/react-native": "^0.65.21",
     "@typescript-eslint/eslint-plugin": "^5.13.0",
     "@typescript-eslint/parser": "^5.13.0",
     "eslint": "^8.10.0",
--- a/package.json
+++ b/package.json
@@ -40,24 +40,24 @@
   },
   "peerDependencies": {
     "react": ">=16.8.6",
-    "react-native": ">=0.59.8"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -40,24 +40,24 @@
...
   "devDependencies": {
-    "@types/jest": "^27.4.0",
+    "@types/jest": "^27.4.1",
     "@types/react": "^17.0.39",
-    "@types/react-native": "^0.66.15",
-    "@typescript-eslint/eslint-plugin": "^5.10.2",
-    "@typescript-eslint/parser": "^5.10.2",
-    "eslint": "^8.8.0",
-    "eslint-plugin-jest": "^26.1.0",
+    "@types/react-native": "^0.67.1",
```
AlexBenny: I'm not an expert of typescript so probably I miss something. I'm a bit confused by this version number.
If we support RN 0.65+ I would expect the types <0.65 (here it's ^0.67).
I've tried to find a place where I can see correspondence between types and related libs but I haven't found anything.

AlexBenny (commented):
Looks good.
Only, I'm a bit confused by the version number of the types for RN, which is greater than the version of RN. Is it ok?

AlexBenny (approved):
👍 👍

PR #162, opened 2022-04-11
Fix import of non-published NSDictionary helpers from SnowplowTracker on iOS (close #161)

### LINKED ISSUE #161: Fix import of non-published NSDictionary helpers from SnowplowTracker on iOS
The RN tracker imports a non-exported file, `NSDictionary+SP_TypeMethods.h` from the iOS SnowplowTracker. The problem only shows up when `use_frameworks! :linkage => :static` is used in the Podfile.

The import should be removed and used methods included in the RN tracker.

matus-tomlein:
This PR addresses issue #161 and removes imports of internal header files (`NSDictinoary+SP_TypeMethods.h` and `SPUtilities.h`) from React Native tracker ObjC code.

I added a `NSDictionary-RNSP_TypeMethods` header and implementation to the RN tracker with the methods that were used. I changed the prefix of the file and methods since they conflicted with the original ones.

Also added a `use_frameworks! :linkage => :static` directive to the Podfile in DemoAppTV which raises errors in case this happens. I didn't add it to the DemoApp project because that uses Flipper and is not compatible with it. But we should be notified of such issues in the future when building the DemoAppTV project.


AlexBenny (approved):
LGTM! 👍 👍

PR #163, opened 2022-04-11
Release/1.2.1

matus-tomlein:
This patch release fixes a bug which imported non-published APIs from the Snowplow iOS tracker and caused errors during compilation.

**Bug Fixes**

Fix import of non-published NSDictionary helpers from SnowplowTracker on iOS (#161)


AlexBenny (approved):
LGTM 🚢

PR #171, opened 2022-09-13
Add anonymous tracking features (close #168)

### LINKED ISSUE #168: Add anonymous tracking features
Anonymous tracking is a feature currently implemented in the JavaScript, iOS, and Android trackers that enables anonymising various user and session identifiers. It affects the following user and session identifiers:

1. Client-side user identifiers: `domain_userid` (equivalent to `client_session.userId` in session context). On mobile, there are also the IDFA identifiers in the `mobile_context` platform context entity.
2. Client-side session identifiers: `domain_sessionid` (equivalent to `client_session.sessionId` in session context)
3. Server-side user identifiers: `network_userid` and `user_ipaddress`

There are several levels to the anonymisation depending on which of the three categories are affected:

## 1. Full client-side anonymisation

In this case, both the client-side user identifiers as well as the client-side session identifiers are anonymised.

### Current options in mobile trackers

One could disable tracking both the `client_session` and `mobile_context` context entities:

```
trackerConfig.sessionContext = false
trackerConfig.platformContext = false
```

### TODO in RN tracker

Disabling the platform context altogether may be a bit wasteful as there is other information (e.g., memory usage, carrier, OS) that are not user or session identifiers (although they could be used for fingerprinting). We could provide an option to only disable the IDFA identifiers in the platform context:

```
trackerConfig.userAnonymisation = true
```

This could set them to null (they are not required in the platform context schema).

## 2. Client-side anonymisation with session tracking

In this case, client-side user identifiers should be disabled but session identifiers should still be tracked.

### Current options in mobile trackers

This is currently not possible in the mobile trackers.

### TODO in RN tracker

We could provide an option to enable session context while disabling all user identifiers (both the ones in `client_session` and `mobile_context` context entities):

```
trackerConfig.sessionContext = true
trackerConfig.userAnonymisation = true
```

The `userId` in `client_session` is a required property, so this could always set it to `00000000-0000-0000-0000-000000000000`.

## 3. Server-side anonymisation

This disables setting the server-side cookies with `network_userid` and also anonymizes the user IP address.

### Current options in mobile trackers

This is currently not possible in the mobile trackers.

### TODO in RN tracker

We could add an option to the tracker configuration:

```
trackerConfig.serverAnonymisation = true
```

This would set the `SP-Anonymous: *` HTTP header for requests going to the Collector. The collector will take care of not setting the request headers and anonymizing the IP address.

matus-tomlein:
Exposes the API from the mobile trackers to enable anonymous tracking features (issue #168).

Anonymous tracking on mobile trackers is provided by two configurations:

- `TrackerConfiguration.userAnonymisation` anonymises user identifiers in the `client_session` and `mobile_context` context entities. The `userId` in `client_session` is replaced with a null UUID and IDFA attributes in the platform context are removed.
- `EmitterConfiguration.serverAnonymisation` anonymises server-assigned identifiers. In particular, these are the `network_userid` and `user_ipaddress`.

This PR exposes the configuration options from the mobile trackers.

### DIFF
--- a/DemoApp/App.js
+++ b/DemoApp/App.js
@@ -116,6 +116,23 @@ const App = () => {
     },
   );
 
+  const anonymousTracker = createTracker(
+    'sp_anon',
+    {
+      endpoint: 'placeholder',
+    },
+    {
+      trackerConfig: {
+        screenViewAutotracking: false, // for tests predictability
+        installAutotracking: false,
+        userAnonymisation: true,
+      },
+      emitterConfig: {
+        serverAnonymisation: true,
+      },
+    },
+  );
+
   const onPressTrackScreenViewEvent = () => {
     tracker.trackScreenViewEvent({name: 'onlyRequired'});
     tracker.trackScreenViewEvent({
@@ -298,6 +315,14 @@ const App = () => {
     });
   };
 
+  const onPressTestAnonymousTracker = () => {
+    anonymousTracker.trackScreenViewEvent({name: 'fromAnonymousTracker'});
+    anonymousTracker.trackStructuredEvent({
+      category: 'AnonymousTracker',
+      action: 'trackAnonymous',
+    });
+  };
+
   const onPressPlayGC = async () => {
     try {
       await tracker.removeGlobalContexts('testTag');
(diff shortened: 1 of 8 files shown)


DemoApp/App.js
```diff
@@ -116,6 +116,23 @@ const App = () => {
     },
   );
 
+  const anonymousTracker = createTracker(
+    'sp_anon',
+    {
+      endpoint: 'http://10.0.2.2:9090',
```
adatzer: nit -  just for consistency, to switch to `'placeholder'` for endpoint

adatzer (approved):
Looks great!!

DemoApp/App.js
matus-tomlein: Fixed, thanks!

PR #173, opened 2022-09-21
Upgrade internal and demo app dependencies (close #172)

### LINKED ISSUE #172: Upgrade internal and demo app dependencies
Upgrade dependencies in the demo apps and target the Android SDK 31.

matus-tomlein:
Addresses issue #172.

1. Upgrades dependencies in the demo apps
2. Targets Android SDK v31 in the demo apps

There are a lot of bumped dependencies. Unfortunately, I wasn't able to isolate the changes from each other as they are quite interlinked.

The Detox CI build still targets Android API 28. I wasn't able to update this as there is an [open issue in Detox](https://github.com/wix/Detox/issues/2612) which causes the CI actions to fail with newer SDKs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Setup Node
       uses: actions/setup-node@v2
       with:
-        node-version: 14
+        node-version: 16
 
     - name: Install node modules
       run: npm ci
@@ -45,7 +45,7 @@ jobs:
     - name: Setup Node
       uses: actions/setup-node@v2
       with:
-        node-version: 14
+        node-version: 16
 
     - name: Build tracker dist files
       run: |
@@ -68,7 +68,7 @@ jobs:
 
   build_ios:
     needs: ["test"]
-    runs-on: macos-10.15
+    runs-on: macos-latest
 
     steps:
     - name: Checkout
@@ -77,7 +77,7 @@ jobs:
     - name: Setup Node
       uses: actions/setup-node@v2
       with:
-        node-version: 14
+        node-version: 16
 
     - name: Build tracker dist files
       run: |
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   e2e:
-    runs-on: macos-10.15
+    runs-on: macos-latest
     timeout-minutes: 30
 
     steps:
@@ -16,13 +16,13 @@ jobs:
     - name: Setup Node
       uses: actions/setup-node@v2
       with:
-        node-version: 14
+        node-version: 16
 
     - name: Setup Java
-      uses: actions/setup-java@v2
+      uses: actions/setup-java@v3
       with:
-        distribution: 'adopt'
-        java-version: '8'
+        distribution: 'zulu'
+        java-version: '11'
 
     # -- Micro --
     - name: Cache Micro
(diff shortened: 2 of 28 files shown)


.github/workflows/build.yml
```diff
@@ -68,7 +68,7 @@ jobs:
 
   build_ios:
     needs: ["test"]
-    runs-on: macos-10.15
+    runs-on: macos-latest
```
adatzer, line 71: would it make sense to pin to `macos-11` instead of `latest`?
(also applies for `deploy.yml`)

.github/workflows/e2e-android.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   e2e:
-    runs-on: macos-10.15
+    runs-on: macos-latest
```
adatzer, line 8: here too about pinning macos version

.github/workflows/e2e-ios.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 jobs:
   e2e:
-    runs-on: macos-10.15
+    runs-on: macos-latest
```
adatzer, line 8: here too about pinning macos version

DemoApp/android/gradle.properties
```diff
@@ -10,7 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx10248m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
adatzer, line 13: out of curiosity, why is this change needed?

adatzer (commented):
Looks great!
We could also update Micro to latest v1.3.3!

.github/workflows/build.yml
matus-tomlein: I don't have a strong opinion which one to use, changed it to `macos-11`.

DemoApp/android/gradle.properties
matus-tomlein: It was running out of memory sometimes when running the Detox tests. It didn't happen all the time but it made the tests more flaky, so I thought it'd be better just to increase the JVM limit to 2GB.

matus-tomlein:
Thanks @adatzer for the review and comments! Updated the Micro review as well.

.github/workflows/build.yml
matus-tomlein: Unfortunately, when I changed it to `macos-11`, the Android e2e CI action started to fail. Detox couldn't connect to the Android emulator. The same was happening when I used `macos-12`. Couldn't figure out why, it seems like a house of cards 😢  I had a similar problem with Detox last week, it was doing the same thing when I tried to build with Android SDK 31 or 32. After a couple of days debugging, I found out that it is an open issue in Detox. I think we'll just need to keep it like this and wait for a more stable version of Detox.

adatzer (approved):
Only nits left:
1. Switching to latest also [in this line](https://github.com/snowplow/snowplow-react-native-tracker/blob/66eb1dfa1f860b371447276fbc2b9c3f68ce8593/.github/workflows/deploy.yml#L110)
2. Micro update also [here](https://github.com/snowplow/snowplow-react-native-tracker/blob/66eb1dfa1f860b371447276fbc2b9c3f68ce8593/DemoApp/bin/docker-run-micro.js#L19)

Looks great!!

PR #175, opened 2022-09-23
Remove circular imports of demo app dendencies when running in demo apps (close #174)

### LINKED ISSUE #174: Remove circular imports of demo app dependencies when running in demo apps
Currently, when one runs `yarn` inside the `DemoApp` or `DemoAppTV` folders, the whole parent folder is copied to satisfy the tracker dependency. However, this also includes the demo app themselves. This can result in ever increasing depth of recursive dependencies inside the demo apps.

The `npm run bootstrap` script in the tracker package handles this problem since it removes the `node_modules` in the `DemoApp` before installing. However, it doesn't consider the `DemoAppTV` app and also doesn't help if one runs `yarn` inside the demo app folder directly.

matus-tomlein:
Issue #174 

# Problem

Currently, when one runs yarn inside the DemoApp or DemoAppTV folders, the whole parent folder is copied to satisfy the tracker dependency. However, this also includes the demo app themselves. This can result in ever increasing depth of recursive dependencies inside the demo apps.

The npm run bootstrap script in the tracker package handles this problem since it removes the node_modules in the DemoApp before installing. However, it doesn't consider the DemoAppTV app and also doesn't help if one runs yarn inside the demo app folder directly.

# Change

This PR adds a postinstall script to the demo apps that removes the `node_modules` folders in the copied tracker library.

# Failing tests

The tests are failing because they use an outdated macOS version for building. However, the PR #173 contains passing tests including this change.

### DIFF
--- a/DemoApp/package.json
+++ b/DemoApp/package.json
@@ -24,7 +24,8 @@
     "e2e:android": "run-p --race start e2e:test:android",
     "e2e:ios": "run-p --race start e2e:test:ios",
     "e2e:android:micro": "run-s -c micro:run e2e:android micro:stop",
-    "e2e:ios:micro": "run-s -c micro:run e2e:ios micro:stop"
+    "e2e:ios:micro": "run-s -c micro:run e2e:ios micro:stop",
+    "postinstall": "rimraf node_modules/@snowplow/react-native-tracker/{.git,node_modules,DemoApp,DemoAppTV}"
   },
   "dependencies": {
     "@jest/create-cache-key-function": "^27.0.2",
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,7 @@
 # Demo app
 #
 DemoApp/
+DemoAppTV/
 
 # Local testing workflow
 #
--- a/DemoAppTV/package.json
+++ b/DemoAppTV/package.json
@@ -8,7 +8,8 @@
     "start": "react-native start",
     "test": "jest",
     "lint": "eslint .",
-    "pods": "cd ios && rimraf Pods && pod install"
+    "pods": "cd ios && rimraf Pods && pod install",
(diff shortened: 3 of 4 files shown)


DemoApp/package.json
```diff
@@ -24,7 +24,8 @@
     "e2e:android": "run-p --race start e2e:test:android",
     "e2e:ios": "run-p --race start e2e:test:ios",
     "e2e:android:micro": "run-s -c micro:run e2e:android micro:stop",
-    "e2e:ios:micro": "run-s -c micro:run e2e:ios micro:stop"
+    "e2e:ios:micro": "run-s -c micro:run e2e:ios micro:stop",
+    "postinstall": "rm -rf node_modules/@snowplow/react-native-tracker/{.git,node_modules,DemoApp,DemoAppTV}"
```
adatzer: I did not know that `yarn` [ignores `files`](https://github.com/yarnpkg/yarn/issues/2822) in package.json for a `file:..` dependency. It does not seem to "install" though, but appears to just copy the tree. 

An alternative could be using `npm pack` (also) in bootstrap script and use the generated `file: ../snowplow-react-native-tracker-1.2.1.tgz` as the tracker dependency in demo apps? What do you think?

If we still go this way, could we use `rimraf` instead of `rm -rf` for an OS/shell independent way?

matus-tomlein: I like the idea to `npm pack`. I tried it out and it worked but was running into issues in CI. Checking the package checksum was failing as yarn was trying to match it to the checksum of the published version (at least that's my understanding). Even when ignoring the checksum check, yarn was producing more errors.

So I decided to go back to the previous approach for now, but used `rimraf` instead of `rm -rf` as you suggested.

adatzer (approved):
LGTM!

PR #176, opened 2022-09-26
Add interface to subscribe for events tracked in Web views (close #169)

### LINKED ISSUE #169: Add interface to subscribe for events tracked in Web views
Add an option to subscribe to events tracked using the [Snowplow WebView tracker](https://github.com/snowplow-incubator/snowplow-webview-tracker) inside Web views. When messages with the events are received, they should be parsed and tracked using the tracker.

matus-tomlein:
Issue #169 

Enables events to be tracked from Web views (implemented using the [react-native-webview](https://github.com/react-native-webview/react-native-webview) component) when the Web view app uses the [WebView tracker](https://github.com/snowplow-incubator/snowplow-webview-tracker).

The communication between Web view app and the React native app is through the `window.ReactNativeInterface` object. The WebView tracker sends messages with tracked events to the object. In the React Native app, one needs to subscribe to the messages by setting the `onMessage` callback in the Web view:

```js
<WebView
   onMessage={getWebViewCallback()} ... />
```

The `getWebViewCallback` function is provided the React native tracker and creates a callback that processes the messages from the WebView tracker and tracks the passed events.

I opened a [PR in the WebView tracker](https://github.com/snowplow-incubator/snowplow-webview-tracker/pull/7) to add support for the `ReactNativeInterface`

## Demo apps

I updated the React Native demo app to add a Web view at the bottom of the app. One needs to set the `webViewEndpoint` variable in the `App.js` file to  to the URL of the app to be shown in the Web view.

There is an exapmle web app in the WebView tracker that can be used to track a few events in [this PR](https://github.com/snowplow-incubator/snowplow-webview-tracker/pull/8).

### DIFF
--- a/DemoApp/App.js
+++ b/DemoApp/App.js
@@ -16,15 +16,27 @@ import {
   useColorScheme,
   View,
   Button,
+  Dimensions,
 } from 'react-native';
+import {WebView} from 'react-native-webview';
 
 import {Colors, Header} from 'react-native/Libraries/NewAppScreen';
 
 import {
   createTracker,
   removeTracker,
-  // removeAllTrackers,
+  getWebViewCallback,
 } from '@snowplow/react-native-tracker';
+
+/**
+ * URI of the Snowplow collector (e.g., Micro, Mini, or BDP) to send events to
+ */
+const collectorEndpoint = 'http://0.0.0.0:9090';
+
+/**
+ * URI of a website to load in the webview component
+ */
+const webViewEndpoint = '';
 
 const Section = ({children, title}) => {
   const isDarkMode = useColorScheme() === 'dark';
@@ -62,7 +74,7 @@ const App = () => {
   const tracker = createTracker(
     'sp1',
     {
-      endpoint: 'placeholder',
+      endpoint: collectorEndpoint,
     },
     {
       trackerConfig: {
@@ -106,7 +118,7 @@ const App = () => {
   const secTracker = createTracker(
     'sp2',
     {
-      endpoint: 'placeholder',
+      endpoint: collectorEndpoint,
     },
     {
       trackerConfig: {
@@ -119,7 +131,7 @@ const App = () => {
   const anonymousTracker = createTracker(
     'sp_anon',
     {
-      endpoint: 'placeholder',
+      endpoint: collectorEndpoint,
     },
     {
       trackerConfig: {
@@ -489,6 +501,21 @@ const App = () => {
               accessibilityLabel="testRemove"
             />
           </Section>
+          <Section title="Web view">
+            {webViewEndpoint ? (
+              <WebView
+                onMessage={getWebViewCallback()}
+                source={{uri: webViewEndpoint}}
+                style={{
+                  height: 400,
+                  width:
+                    Dimensions.get('window').width -
+                    styles.sectionContainer.paddingHorizontal,
+                }}
+              />
+            ) : null}
+          </Section>
+          <Section title="Last section"></Section>
         </View>
       </ScrollView>
     </SafeAreaView>
--- a/DemoApp/App.js
+++ b/DemoApp/App.js
@@ -31,7 +31,7 @@ import {
 /**
  * URI of the Snowplow collector (e.g., Micro, Mini, or BDP) to send events to
  */
-const collectorEndpoint = 'http://0.0.0.0:9090';
+const collectorEndpoint = 'placeholder';
 
 /**
  * URI of a website to load in the webview component
@@ -515,7 +515,6 @@ const App = () => {
               />
             ) : null}
           </Section>
-          <Section title="Last section"></Section>
         </View>
       </ScrollView>
     </SafeAreaView>
--- a/android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
+++ b/android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
@@ -137,7 +137,7 @@ public class RNSnowplowTrackerModule extends ReactContextBaseJavaModule {
                               Promise promise) {
         try {
             String namespace = details.getString("tracker");
-            TrackerController trackerController = Snowplow.getTracker(namespace);
+            TrackerController trackerController = getTracker(namespace);
 
             promise.resolve(Snowplow.removeTracker(trackerController));
 
@@ -165,7 +165,7 @@ public class RNSnowplowTrackerModule extends ReactContextBaseJavaModule {
             ReadableMap argmap = details.getMap("eventData");
             ReadableArray contexts = details.getArray("contexts");
 
-            TrackerController trackerController = Snowplow.getTracker(namespace);
+            TrackerController trackerController = getTracker(namespace);
 
             SelfDescribingJson sdj = EventUtil.createSelfDescribingJson(argmap);
             SelfDescribing event = new SelfDescribing(sdj);
@@ -189,7 +189,7 @@ public class RNSnowplowTrackerModule extends ReactContextBaseJavaModule {
             ReadableMap argmap = details.getMap("eventData");
             ReadableArray contexts = details.getArray("contexts");
 
-            TrackerController trackerController = Snowplow.getTracker(namespace);
+            TrackerController trackerController = getTracker(namespace);
 
             Structured event = EventUtil.createStructuredEvent(argmap);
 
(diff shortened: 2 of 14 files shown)


DemoApp/App.js
```diff
@@ -16,16 +16,28 @@ import {
...
   createTracker,
   removeTracker,
-  // removeAllTrackers,
+  getWebViewCallback,
 } from '@snowplow/react-native-tracker';
 
+/**
+ * URI of the Snowplow collector (e.g., Micro, Mini, or BDP) to send events to
+ */
+const collectorEndpoint = 'http://0.0.0.0:9090';
```
adatzer: nit - `placeholder`

DemoApp/App.js
```diff
@@ -489,6 +501,21 @@ const App = () => {
...
+                style={{
+                  height: 400,
+                  width:
+                    Dimensions.get('window').width -
+                    styles.sectionContainer.paddingHorizontal,
+                }}
+              />
+            ) : null}
+          </Section>
+          <Section title="Last section"></Section>
```
adatzer: is this last section meant to be empty?

android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
```diff
@@ -137,7 +137,7 @@ public void removeTracker(ReadableMap details,
                               Promise promise) {
         try {
             String namespace = details.getString("tracker");
-            TrackerController trackerController = Snowplow.getTracker(namespace);
+            TrackerController trackerController = getTracker(namespace);
```
adatzer, line 140: if i understand this correctly, the change is that if namespace is null, we revert to default tracker, right?
Just wanted to confirm: if `tracker` is null, will `getString` return a null namespace as well?

ios/RNSnowplowTracker.m
```diff
@@ -123,7 +123,7 @@ @implementation RNSnowplowTracker
                   resolver:(RCTPromiseResolveBlock)resolve
                   rejecter:(RCTPromiseRejectBlock)reject) {
     NSString *namespace = [details objectForKey:@"tracker"];
-    id<SPTrackerController> trackerController = [SPSnowplow trackerByNamespace:namespace];
+    id<SPTrackerController> trackerController = [self trackerByNamespace:namespace];
```
adatzer, line 126: same question to confirm for null `tracker` value: is the previous line(125) returning `NSNull` for namespace in that case?

src/tracker.ts
```diff
@@ -127,7 +127,7 @@ function trackStructuredEvent(
  * @returns {Promise}
  */
 function trackPageViewEvent(
-  namespace: string,
+  namespace: string | null,
```
adatzer, line 130: thinking about it feels a bit surprising that `null` means default tracker.. but i cannot pinpoint exactly why! Do you think it is worth to consider alternatives? 
Also, if we still go this way, we could create a tracker namespace type to use in all functions and not only to those related to webview tracking?

src/webViewInterface.ts
```diff
@@ -0,0 +1,100 @@
...
+  trackSelfDescribingEvent,
+  trackStructuredEvent,
+} from './tracker';
+import type { ScreenViewProps, SelfDescribing, StructuredProps } from './types';
+
+interface FullPageViewEvent {
+  title?: string | null;
+  url?: string;
+  referrer?: string;
+}
```
adatzer: why not use `PageViewProps` [type](https://github.com/snowplow/snowplow-react-native-tracker/blob/b76c84ce32ceffa58a6bf1bdc8ceaaa9eb8b78d4/src/types.ts#L392) here?

src/webViewInterface.ts
```diff
@@ -0,0 +1,100 @@
...
+
+type WebViewMessage = {
```
adatzer: we could also move all type definitiions to `types.ts` ?

adatzer (commented):
It is great to offer this functionality in RN tracker!

DemoApp/App.js
matus-tomlein: Thanks, fixed!

DemoApp/App.js
matus-tomlein: Good catch! This shouldn't be here, I removed it.

android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
matus-tomlein: Yes, that is correct. The WebView tracker has the ability to track with the default tracker (without specifying a concrete namespace) so this is meant to facilitate that.

The important thing to note is that the change is only in the internals of the tracker. The public API of the React Native tracker doesn't change and still requires events to be tracked on a tracker instance.

ios/RNSnowplowTracker.m
matus-tomlein: Yes, same here, the namespace will be `NSNull` in that case.

src/webViewInterface.ts
matus-tomlein: The `PageViewProps` type has different attributes – `pageTitle` instead `title`, and `pageUrl` instead of `url`.

src/webViewInterface.ts
matus-tomlein: I'm happy to do that but just want to point out that this is an internal type, it shouldn't be exposed anywhere. I didn't add it to `types.ts` because I thought that contains the public types. But if that makes more sense, I can move it over there.

src/tracker.ts
matus-tomlein: I can think of one other alternative – to fetch the namespace of the default tracker from the native tracker in the TypeScript code and then use that namespace to track the events. We can change it to do that – it would require a new API to get the default tracker namespace in `RNSnowplowTrackerModule` and `RNSnowplowTracker` and exchanging two messages instead of one between the TS and native code. But it also seems like a good solution to me.

> we could create a tracker namespace type to use in all functions

I'm not sure I understand what you mean here?

android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
adatzer: Thank you for explaining @matus-tomlein ! This is clear!

src/tracker.ts
adatzer: Sorry i was not clear. What i meant is that currently we have `trackPageViewEvent`, `trackStructuredEvent`, `trackSelfDescribingEvent` and `trackScreenViewEvent` accepting

```
namespace: string | null
```

and the rest of the functions (e.g. `trackTimingEvent`) accepting

```
namespace: string
```

Even though those functions cannot be called in the context of webviews, it may be a bit clearer to have the same type for this property. I was just suggesting a type definition to be applied for all. Not a blocking issue at all though, as it may not be worth it.

src/webViewInterface.ts
adatzer: Missed this, you are right! Would it make sense to add `WebView` as prefix instead of `Full` ?

matus-tomlein: Yes, that sounds better, renamed it to `WebViewPageViewEvent` and also moved to `types.ts`.

src/webViewInterface.ts
matus-tomlein: On second thought, I moved it to `types.ts`, added comments to indicate that it is an internal type.

src/tracker.ts
matus-tomlein: Oh, I see, that's a good point, it's better to be consistent in all the internal track methods. I changed the others as well.

src/webViewInterface.ts
```diff
@@ -0,0 +1,89 @@
+import {
+  trackPageViewEvent,
+  trackScreenViewEvent,
+  trackSelfDescribingEvent,
+  trackStructuredEvent,
+} from "./tracker";
```
adatzer: nit to please eslint - single quotes

src/webViewInterface.ts
```diff
@@ -0,0 +1,89 @@
...
+import type {
+  ScreenViewProps,
+  SelfDescribing,
+  StructuredProps,
+  WebViewPageViewEvent,
+  WebViewMessage,
+} from "./types";
```
adatzer: nit - single quotes

adatzer (approved):
Looks great to me!!

PR #177, opened 2022-09-28
Add option to add custom headers (close #165)

### LINKED ISSUE #165: Add option to add custom headers
**Is your feature request related to a problem? Please describe.**
Some users want to place the Snowplow collector behind an authenticated gateway which requires custom headers to be sent with the request. Currently, the tracker is unable to send requests successfully to this as there is no way to easily add custom headers to the request.

**Describe the solution you'd like**
Create a configuration option that allows for custom headers to be added.
The added custom headers would configure the native trackers wrapped by the react native tracker.

matus-tomlein:
Issue #165 

Enables adding custom HTTP request headers to the collector requests using the `NetworkConfiguration.requestHeaders` option. The configuration is set by passing a dictionary of the headers, similar to the iOS tracker:

```js
networkConfig.requestHeaders = {
  "sp": "anonymous"
};
```

On iOS, the implementation is straightforward as it already provides the same configuration option in `NetworkConfiguration`.

However, the Android tracker doesn't expose an option to set custom headers in configuration. I had to do it by setting a custom HTTP client (`OkHttpClient`) which intercepts requests and adds the headers to them. We might change this in case we expose a configuration option for setting headers in the Android tracker.

### DIFF
--- a/android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
+++ b/android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
@@ -8,12 +8,12 @@ import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 
+import com.facebook.react.bridge.ReadableMapKeySetIterator;
 import com.snowplowanalytics.snowplow.Snowplow;
 import com.snowplowanalytics.snowplow.configuration.Configuration;
 import com.snowplowanalytics.snowplow.configuration.EmitterConfiguration;
 import com.snowplowanalytics.snowplow.configuration.GdprConfiguration;
 import com.snowplowanalytics.snowplow.configuration.NetworkConfiguration;
-import com.snowplowanalytics.snowplow.configuration.RemoteConfiguration;
 import com.snowplowanalytics.snowplow.configuration.SessionConfiguration;
 import com.snowplowanalytics.snowplow.configuration.TrackerConfiguration;
 import com.snowplowanalytics.snowplow.configuration.SubjectConfiguration;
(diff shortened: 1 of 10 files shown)


src/types.ts
```diff
@@ -82,6 +82,11 @@ export interface NetworkConfiguration {
    * @defaultValue 'post'
    */
   method?: HttpMethod;
+
+  /**
+   * Custom headers for HTTP requests to the Collector.
+   */
+  requestHeaders?: [string: string]
```
igneel64: Nit:
```ts
  /**
   * Custom headers for HTTP requests to the Collector.
   * E.g. requestHeaders: { 'X-Snowplow': 'value' }
   */
  requestHeaders?: [name: string]
```

src/types.ts
```diff
@@ -82,6 +82,11 @@ export interface NetworkConfiguration {
    * @defaultValue 'post'
    */
   method?: HttpMethod;
+
+  /**
+   * Custom headers for HTTP requests to the Collector.
+   */
+  requestHeaders?: [name: string]
```
adatzer: how about matching (option name, type and defaultValue) [the JavaScript tracker](https://github.com/snowplow/snowplow-javascript-tracker/blob/dc8fa2a1df02e74bbb1ee5c313c2b64907b152ea/libraries/browser-tracker-core/src/tracker/types.ts#L219-L223) here?

android/src/main/java/com/snowplowanalytics/react/tracker/RNSnowplowTrackerModule.java
```diff
@@ -73,6 +76,26 @@ public void createTracker(ReadableMap argmap,
             } else {
                 networkConfiguration = new NetworkConfiguration(networkConfig.getString("endpoint"));
             }
+            if (networkConfig.hasKey("requestHeaders") && !networkConfig.isNull("requestHeaders")) {
+                ReadableMap requestHeaders = networkConfig.getMap("requestHeaders");
+                if (requestHeaders != null) {
+                    OkHttpClient client = new OkHttpClient.Builder()
```
adatzer, line 86: this is nice workaround! still, it will be great when the Android tracker exposes this option as well..

igneel64 (approved):

src/types.ts
matus-tomlein: I changed the type to match the JavaScript tracker but I think we should keep the name `requestHeaders` to match the `NetworkConfiguration` in the iOS tracker. So far the RN configuration exactly matches the mobile trackers configuration and I think we should keep that.

adatzer (approved):
LGTM!!

PR #179, opened 2022-09-29
Enable setting custom path for POST requests (close #178)

### LINKED ISSUE #178: Enable setting custom path for POST requests
Add the configuration to set custom path for POST requests to the collector. The default is `com.snowplowanalytics.snowplow/tp2`.

The configuration should be in `NetworkConfiguration.customPostPath`.

matus-tomlein:
Issue #178 

Adds the optional `customPostPath` setting to the `NetworkConfiguration` that changes the default path for POST requests to the collector.

The configuration option was already provided by the mobile native trackers, so this PR just exposes the configuration in the RN tracker configuration.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -87,7 +87,7 @@ export interface NetworkConfiguration {
    * A custom path which will be added to the endpoint URL to specify the
    * complete URL of the collector when paired with the POST method.
    * 
-   * The default is `com.snowplowanalytics.snowplow/tp2`.
+   * @defaultValue `com.snowplowanalytics.snowplow/tp2`.
    */
    customPostPath?: string;
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -86,6 +86,8 @@ export interface NetworkConfiguration {
   /**
    * A custom path which will be added to the endpoint URL to specify the
    * complete URL of the collector when paired with the POST method.
+   * 
+   * The default is `com.snowplowanalytics.snowplow/tp2`.
    */
    customPostPath?: string;
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -82,6 +82,12 @@ export interface NetworkConfiguration {
    * @defaultValue 'post'
    */
   method?: HttpMethod;
+
+  /**
+   * A custom path which will be added to the endpoint URL to specify the
+   * complete URL of the collector when paired with the POST method.
+   */
+   customPostPath?: string;
 }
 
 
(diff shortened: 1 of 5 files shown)


tests/__tests__/configurations.test.ts
```diff
@@ -207,7 +207,7 @@ describe('test initValidate resolves', () => {
   test('test with all defaults', async () => {
     const testConfig = {
       namespace: 'sp1',
-      networkConfig: {endpoint: 'test', method: 'post'},
+      networkConfig: {endpoint: 'test', method: 'post', customPostPath: '/'},
```
adatzer: nit - just because of this test's title to use `/com.snowplowanalytics.snowplow/tp2`?

adatzer (commented):
Looks great! 
I initially got a bit confused whether a leading slash is needed or not (not for Android, but i haven't tested on iOS, so i can't confirm which is the case). We could actually test this on CI, modifying `micro.conf` [here](https://github.com/snowplow/snowplow-react-native-tracker/blob/b76c84ce32ceffa58a6bf1bdc8ceaaa9eb8b78d4/DemoApp/tests/micro/micro.conf#L26-L30). What do you think?

igneel64 (approved):

tests/__tests__/configurations.test.ts
matus-tomlein: Good point! I changed it to `com.snowplowanalytics.snowplow/tp2` because the leading slash is added by default. Also changed in other places to remove the leading slash.

matus-tomlein:
Thanks Ada! Integration tests might be good but I don't think it's really worth the effort here because we only expose the configuration options of the mobile trackers. I think there is a better case to put the integration tests on the mobile trackers side (they already have tests for this although not with Micro).

src/types.ts
```diff
@@ -82,6 +82,14 @@ export interface NetworkConfiguration {
    * @defaultValue 'post'
    */
   method?: HttpMethod;
+
+  /**
+   * A custom path which will be added to the endpoint URL to specify the
+   * complete URL of the collector when paired with the POST method.
+   * 
+   * The default is `com.snowplowanalytics.snowplow/tp2`.
```
adatzer: nit - ` * @defaultValue 'com.snowplowanalytics.snowplow/tp2'` could be used too

adatzer (approved):
LGTM!!

PR #180, opened 2022-10-03
Release/1.3.0

matus-tomlein:
This release upgrades the underlying mobile trackers to version 4 and introduces new features such as anonymous tracking and Web view tracking. It also adds a couple of new network configuration options (custom request headers and custom POST path).

* Upgrade Snowplow mobile trackers to v4 (#167)
* Add anonymous tracking features (#168)
* Add interface to subscribe for events tracked in Web views (#169)
* Add option to add custom headers (#165)
* Enable setting custom path for POST requests (#178)
* Upgrade internal and demo app dependencies (#172)
* Remove circular imports of demo app dendencies when running in demo apps (#174)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,13 @@
+Version 1.3.0 (2022-10-03)
+--------------------------
+Upgrade Snowplow mobile trackers to v4 (#167)
+Add anonymous tracking features (#168)
+Add interface to subscribe for events tracked in Web views (#169)
+Add option to add custom headers (#165)
+Enable setting custom path for POST requests (#178)
+Upgrade internal and demo app dependencies (#172)
+Remove circular imports of demo app dependencies when running in demo apps (#174)
+
 Version 1.2.1 (2022-04-12)
 --------------------------
 Fix import of non-published NSDictionary helpers from SnowplowTracker on iOS (#161)
--- a/src/webViewInterface.ts
+++ b/src/webViewInterface.ts
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2020-2022 Snowplow Analytics Ltd. All rights reserved.
+ *
+ * This program is licensed to you under the Apache License Version 2.0,
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+
+'use strict';
+
+import {
+  trackPageViewEvent,
+  trackScreenViewEvent,
+  trackSelfDescribingEvent,
+  trackStructuredEvent,
+} from './tracker';
+import type {
+  ScreenViewProps,
+  SelfDescribing,
+  StructuredProps,
+  WebViewPageViewEvent,
+  WebViewMessage,
+} from './types';
+import { errorHandler } from './utils';
+
+function forEachTracker(
+  trackers: Array<string> | undefined,
+  iterator: (namespace: string | null) => void
+): void {
+  if (trackers && trackers.length > 0) {
+    trackers.forEach(iterator);
+  } else {
+    iterator(null);
+  }
+}
+
+/**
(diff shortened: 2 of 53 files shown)


CHANGELOG
```diff
@@ -1,3 +1,13 @@
+Version 1.3.0 (2022-10-03)
+--------------------------
+Upgrade Snowplow mobile trackers to v4 (#167)
+Add anonymous tracking features (#168)
+Add interface to subscribe for events tracked in Web views (#169)
+Add option to add custom headers (#165)
+Enable setting custom path for POST requests (#178)
+Upgrade internal and demo app dependencies (#172)
+Remove circular imports of demo app dendencies when running in demo apps (#174)
```
adatzer: nit typo- `dependencies` (also in commit message)

adatzer (approved):
Looks great!
Is this PR waiting for the next release of webview tracker to include this [pull request](https://github.com/snowplow-incubator/snowplow-webview-tracker/pull/7)?

src/webViewInterface.ts
```diff
@@ -0,0 +1,89 @@
+import {
```
adatzer, line 16: :no_entry_sign:  The copyright notice is missing in this file. 
:pick:  Also, we could `use strict` like in other ts files.

src/webViewInterface.ts
```diff
@@ -0,0 +1,89 @@
...
+  StructuredProps,
+  WebViewPageViewEvent,
+  WebViewMessage,
+} from './types';
+
+function forEachTracker(
+  trackers: Array<string> | undefined,
+  iterator: (namespace: string | null) => void
+): void {
+  if (trackers) {
```
adatzer: :question: would it make sense to add a check for empty array? what is supposed to happen then? (at the moment the event is not tracked)

src/webViewInterface.ts
```diff
@@ -0,0 +1,89 @@
...
+    const data = JSON.parse(message.nativeEvent.data) as WebViewMessage;
+    switch (data.command) {
+    case 'trackSelfDescribingEvent':
+      forEachTracker(data.trackers, (namespace) => {
+        trackSelfDescribingEvent(
+          namespace,
+            data.event as SelfDescribing,
+            data.context || []
+        ).catch((error) => {
+          console.error(error);
```
adatzer: :pick:  instead of `console.error` the `errorHandler` from `utils.ts` could be used, which does not log on prod. Do you think it'd make sense?

src/webViewInterface.ts
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * Enables tracking events from apps rendered in react-native-webview components.
+ * The apps need to use the Snowplow WebView tracker to track the events.
+ * 
+ * To subscribe for the events, set the `onMessage` attribute:
+ * <WebView onMessage={getWebViewCallback()} ... />
+ * 
+ * @returns Callback to subscribe for events from Web views tracked using the Snowplow WebView tracker.
+ */
+export function getWebViewCallback() {
```
adatzer, line 51: :bulb: Currently, if the tracker namespace does not belong to an initialized tracker, we get an error from native side. 
What if `getWebViewCallback` is made to accept full [trackers](https://github.com/snowplow/snowplow-react-native-tracker/blob/b76c84ce32ceffa58a6bf1bdc8ceaaa9eb8b78d4/src/types.ts#L657) as arguments? e.g. then to be used with something like `onMessage={getWebViewCallback(tracker)}`

On one hand, this brings the question about what happens with the tracker namespaces in the webview message (e.g. ignore them/cross-check/?). On the other hand, this may sound a bit more inline with how the RN tracker currently ensures that a tracker is initialized before it calls any of its methods. 
Not an objection though - just sharing late thoughts :see_no_evil:

adatzer (commented):
After this webview-tracker [PR](https://github.com/snowplow-incubator/snowplow-webview-tracker/pull/13) got a bit carried away with webviews in android. I apologize for changing the review at this point, as i failed to consider these sooner.

src/webViewInterface.ts
matus-tomlein: Good point, fixed!

src/webViewInterface.ts
matus-tomlein: Yeah, I think that makes sense to be sure, added the check.

src/webViewInterface.ts
matus-tomlein: Good catch, thanks!

src/webViewInterface.ts
matus-tomlein: That's a good idea, will keep in mind for the next release!

adatzer (approved):

PR #184, opened 2023-01-23
Add JavaScript tracker core to be used in Expo managed workflow when native trackers are not available

matus-tomlein:
Issue #188 

This PR adds support for the [Expo Go](https://docs.expo.dev/workflow/expo-go/) and managed workflow in the Expo framework in which the tracker is not allowed to access native iOS and Android code – that includes the native mobile trackers.

To work around this limitation, I added a fallback in case the native code is not available. The fallback accesses the tracker core from the JavaScript tracker to track events. I wasn't able to use the browser or NodeJS trackers because of their requirements, only the core was possible to use.

The implementation is quite limited as the more advanced autotracking features are not available. Also, session tracking is not implemented at the moment, but we may provide a simple implementation in another PR. The emitter is very simple as it sends events right after being tracked.

Despite the limitations, I think the limited support is better than nothing. We do not have another alternative solution for Expo Go users so this gives at least something to work with. We may improve it further if user requests come in.

## Documentation: Support for the Expo framework

The React Native tracker supports the [Expo framework](https://expo.dev/). However, the support is different for the bare and the managed workflow.

You may install the tracker using:

```sh
npx expo install @snowplow/react-native-tracker
```

### Expo bare workflow

The tracker works without limitations on the bare workflow. Since the tracker is able to use iOS and Android native code using the mobile trackers, it provides the full functionality of the React Native tracker.

Make sure to rebuild your app after installing the tracker.

### Expo managed workflow and Expo Go

The functionality of the tracker is limited on the managed workflow. This is because the tracker is unable to run native iOS and Android code of the mobile trackers that it builds on. However, basic functionality is provided through a JavaScript-only implementation.

The following functionality is available:

- Manual tracking of all events using the tracker APIs (e.g., using the `trackScreenView()` function).
- Adding context entities to manually tracked events.
- Setting Subject properties to add information to events.
- Having multiple tracker instances with different configuration in the app.

The tracker is missing more advanced features such as:

- Session tracking – the tracker does not automatically add the client session context entity to events currently. However, you may add it yourself.
- Auto tracking features – features like install auto-tracking that track events automatically are not available.
- No support for batching multiple events in POST requests from the emitter – events are sent one per request as soon as they are tracked.


src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+  buildScreenView,
+  buildPageView,
+  buildConsentGranted,
+  buildConsentWithdrawn,
+  buildEcommerceTransaction,
+} from '@snowplow/tracker-core';
+import type { TrackerCore } from '@snowplow/tracker-core';
+
+// Tracker version added to the events
+const trackerVersion = 'rn-1.3.0';
```
adatzer, line 51: since we are introducing a tracker version here, we could also check for it in `deploy.yml` [here as well](https://github.com/snowplow/snowplow-react-native-tracker/blob/20f6000c6e37ae903abb75d7b421a373ef207f65/.github/workflows/deploy.yml#L47-L71)!

src/native.ts
```diff
@@ -14,8 +14,16 @@
...
 
-const { RNSnowplowTracker } = NativeModules;
+const isAvailable = NativeModules.RNSnowplowTracker != null;
+if (!isAvailable) {
+  console.warn(
+    'Unable to access the native iOS/Android Snowplow tracker, a tracker implementation with very limited functionality is used.'
+  );
+}
+
+const RNSnowplowTracker = isAvailable ? NativeModules.RNSnowplowTracker : JSSnowplowTracker;
```
adatzer: Because of the eslint warning at this line `Unsafe assignment of an any value`:
Since in `jsCore.ts` with `JSSnowplowTracker` we are actually implementing the type expected from native modules (and also implemented by `RNSnowplowTracker`), what if we define this explicitly as a type,say `Native`,  that would be covering for both? Besides being explicit, we could use a type assertion (e.g. `NativeModules.RNSnowplowTracker as Native`) to resolve the warning as well. 
What do you think?

src/native.ts
```diff
@@ -14,8 +14,16 @@
 'use strict';
 
 import { NativeModules } from 'react-native';
+import { JSSnowplowTracker } from './jsCore';
 
-const { RNSnowplowTracker } = NativeModules;
+const isAvailable = NativeModules.RNSnowplowTracker != null;
+if (!isAvailable) {
+  console.warn(
```
adatzer: We could also use the `errorHandler` defined [here](https://github.com/snowplow/snowplow-react-native-tracker/blob/20f6000c6e37ae903abb75d7b421a373ef207f65/src/utils.ts#L58) that also checks if in dev before warning. What do you think?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+    tracker.setUserId(configuration.subjectConfig.userId);
+  }
+  if (configuration.subjectConfig?.useragent != null) {
+    tracker.setUseragent(configuration.subjectConfig.useragent);
+  }
+}
+
+function createTracker(
+  configuration: InitTrackerConfiguration,
+  emitCallback?: (e: Payload) => void
```
adatzer, line 138: To confirm the optional argument is needed for testing?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+
+  // update tracker properties to reflect subject and tracker info
+  updateTrackerProperties(tracker, configuration);
+
+  return <Promise<void>>Promise.resolve();
+}
+
+function removeTracker(details: { tracker: string }): Promise<void> {
+  delete trackers[details.tracker];
+  return <Promise<void>>Promise.resolve();
```
adatzer: Could that be `<Promise<boolean>>Promise.resolve(true);` in order for `removeTracker` to have same signature with the corresponding [`RNSnowplowTracker.removeTracker` function](https://github.com/snowplow/snowplow-react-native-tracker/blob/20f6000c6e37ae903abb75d7b421a373ef207f65/src/api.ts#L64) ?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+}
+
+function removeAllTrackers(): Promise<void> {
+  trackers = {};
+  return <Promise<void>>Promise.resolve();
```
adatzer: Same here to match [`RNSnowplowTracker.removeAllTrackers` function signature](https://github.com/snowplow/snowplow-react-native-tracker/blob/20f6000c6e37ae903abb75d7b421a373ef207f65/src/api.ts#L73)?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+
+  for (const key in payload) {
+    if (Object.prototype.hasOwnProperty.call(payload, key)) {
+      stringifiedPayload[key] = String(payload[key]);
+    }
+  }
+  return stringifiedPayload;
+}
+
+function createEmitCallback(configuration: InitTrackerConfiguration): (e: Payload) => void {
```
adatzer: nit - since only `networkConfig` is used in the function, `configuration` could be of just `NetworkConfiguration` type

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+      data: [preparePayload(e)],
+    };
+    const endpoint = configuration.networkConfig.endpoint;
+    const postPath =
+      configuration.networkConfig.customPostPath ??
+      '/com.snowplowanalytics.snowplow/tp2';
+    fetch(endpoint + postPath, {
+      method: 'POST',
+      body: JSON.stringify(postJson),
+      headers: {
```
adatzer, line 86: `method` and `headers` also exist in network configuration. Would it make sense to use them for `fetch` options?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+        'Content-Type': 'application/json',
+      },
+    }).catch(console.error);
```
adatzer: we could also use the `errorHandler` here as well so as those errors don't surface in production environments

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+function forTracker(
+  namespace: string | null,
+  callback: (tracker: Tracker) => void
+): void {
+  const tracker = namespace != null ? trackers[namespace] : Object.values(trackers)[0];
+
+  if (tracker) {
+    callback(tracker);
+  } else {
+    console.error('No such tracker found.');
```
adatzer: could we use `errorHandler` here as well instead?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+      schema:
+        'iglu:com.snowplowanalytics.mobile/message_notification/jsonschema/1-0-0',
+      data: details.eventData,
+    },
+    contexts: details.contexts,
+  });
+}
+
+function removeGlobalContexts(details: { tracker: string; removeTag: string }): Promise<void> {
+  details;
```
adatzer: nit - is this to be removed?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+  return <Promise<void>>Promise.reject('Not implemented');
+}
+
+function addGlobalContexts(details: {
+  tracker: string;
+  addGlobalContext: GlobalContext;
+}): Promise<void> {
+  details;
```
adatzer: same for `details;` here

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+function setUserId(details: { tracker: string; userId: string | null }): Promise<void> {
+  if (details.userId != null) {
+    trackers[details.tracker]?.setUserId(details.userId);
+  }
+  return <Promise<void>>Promise.resolve();
+}
+
+function setNetworkUserId(): Promise<void> {
+  return <Promise<void>>Promise.reject('Not implemented');
+}
```
adatzer, line 405: not necessary at all, just wondering since this is the only not implemented setter, could it be added like the `setDomainUserId`?

src/jsCore.ts
```diff
@@ -0,0 +1,498 @@
...
+function addGlobalContexts(details: {
+  tracker: string;
+  addGlobalContext: GlobalContext;
+}): Promise<void> {
+  details;
+  return <Promise<void>>Promise.reject('Not implemented');
+}
+
+function setUserId(details: { tracker: string; userId: string | null }): Promise<void> {
+  if (details.userId != null) {
```
adatzer: this comment goes to all `set...` functions: There is a difference in how `null` is handled between `JSSnowplowTracker` and `RNSnowplowTracker`. For example [here](https://github.com/snowplow/snowplow-react-native-tracker/blob/20f6000c6e37ae903abb75d7b421a373ef207f65/src/subject.ts#L35). I do see that tracker-core accepts only argument of type e.g. string in this one. What do you think about it? Is it something to consider? I don't see this as a blocker, as feature parity is not a requirement, just wondering.

adatzer (commented):
It is great to support this and very nice way to implement!
Only added some first comments to help me gain some context.

src/jsCore.ts
matus-tomlein: Right, I only added that for testing the created payloads.

src/jsCore.ts
matus-tomlein: Good catch! I fixed the return type.

src/jsCore.ts
matus-tomlein: Fixed, thanks!

src/jsCore.ts
matus-tomlein: Good point, added the version check there too.

src/native.ts
matus-tomlein: I like that idea! I have implemented an interface `Native` for the objects.

src/native.ts
matus-tomlein: Changed to use the `errorHandler`

src/jsCore.ts
matus-tomlein: Using `errorHandler` now too.

src/jsCore.ts
matus-tomlein: Yes, that got forgotten, I have removed it.

src/jsCore.ts
matus-tomlein: Removed.

src/jsCore.ts
matus-tomlein: That's a good point, have implemented the `setNetworkUserId` in the same way as `setDomainUserId`.

src/jsCore.ts
matus-tomlein: Changed to use the `errorHandler`

src/jsCore.ts
matus-tomlein: I have implemented support for the custom headers, but I think we don't need to implement the `GET` method support – we don't recommend using GET anyway and for this minimal implementation, I think we can get away without it.

src/jsCore.ts
matus-tomlein: I think that is okay (unless I misunderstand something) because the `subject.ts` file is called before the `JSSnowplowTracker` is called – it calls the native functions. So the checks for `null` that are there also apply for the `JSSnowplowTracker`.

matus-tomlein:
Thanks a lot for the review and comments @adatzer!

.github/workflows/deploy.yml
```diff
@@ -49,11 +49,13 @@ jobs:
       env:
         IOS_VER_FILEPATH: 'ios/Util/RNTrackerVersion.m'
         ANDR_VER_FILEPATH: 'android/src/main/java/com/snowplowanalytics/react/util/TrackerVersion.java'
+        JS_VER_FILEPATH: 'src/jsCore.ts'
       run: |
         echo ::set-output name=TAG_VERSION::${GITHUB_REF#refs/*/}
         echo "##[set-output name=RN_TRACKER_VERSION;]$(node -p "require('./package.json').version")"
         echo "##[set-output name=RN_IOS_TRACKER_VERSION;]$(cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = @"rn-\(.*\)";/\1/p')"
         echo "##[set-output name=RN_ANDROID_TRACKER_VERSION;]$(cat "${ANDR_VER_FILEPATH}" | sed -n -e 's/^.*RN_TRACKER_VERSION = "rn-\(.*\)";/\1/p')"
+        echo "##[set-output name=RN_JS_TRACKER_VERSION;]$(cat "${JS_VER_FILEPATH}" | sed -n -e 's/^.*trackerVersion = 'rn-\(.*\)';/\1/p')"
```
adatzer: Currently the single quotes are not escaped inside the sed command. We could enclose the sed command in double quotes (safe in this case), but then we'd have to think of escaping those for the outer command :sob: .
Instead we could go with something like:

```
sed -n -e 's/^.*trackerVersion = '\''rn-\(.*\)'\'';/\1/p'
```

Once this is on a release branch we could go for an `rc` release just to check this running as well!

.github/workflows/e2e-android.yml
```diff
@@ -34,7 +34,7 @@ jobs:
 
     - name: Get micro
       if: steps.cache-micro.outputs.cache-hit != 'true'
-      run: curl -o micro.jar -L https://github.com/snowplow-incubator/snowplow-micro/releases/download/micro-1.3.3/snowplow-micro-1.3.3.jar
+      run: curl -o micro.jar -L https://github.com/snowplow-incubator/snowplow-micro/releases/download/micro-1.3.4/snowplow-micro-1.3.4.jar
```
adatzer: [Micro `1.4.0`](https://github.com/snowplow-incubator/snowplow-micro/releases/tag/micro-1.4.0) has just been released!

src/native.ts
```diff
@@ -14,8 +14,157 @@
...
+  ScreenSize,
+  ScreenViewProps,
+  SelfDescribing,
+  StructuredProps,
+  TimingProps,
+} from './types';
+import { errorHandler } from './utils';
 
-const { RNSnowplowTracker } = NativeModules;
+interface Native {
```
adatzer: We could also define this in `types.ts`  as well, so we wouldn't have to import all the required types here.

src/jsCore.ts
adatzer: The check is the opposite though. What i mean is that with native RN tracker it is possible to set e.g. the user id to `null`/back to `null`. However this cannot be done with the `JSSnowplowTracker`?

src/jsCore.ts
```diff
@@ -0,0 +1,509 @@
...
+
+function getBackgroundIndex(): Promise<number> {
+  return <Promise<number>>Promise.reject('Not implemented');
+}
+
+function getForegroundIndex(): Promise<number> {
+  return <Promise<number>>Promise.reject('Not implemented');
+}
+
+const JSSnowplowTracker = {
```
adatzer: nit - what about an `Object.freeze` here?

src/jsCore.ts
```diff
@@ -0,0 +1,509 @@
...
+    }
+  }
+  return stringifiedPayload;
+}
+
+function createEmitCallback(networkConfig: NetworkConfiguration): (e: Payload) => void {
+  return (e: Payload) => {
+    const postJson = {
+      schema:
+        'iglu:com.snowplowanalytics.snowplow/payload_data/jsonschema/1-0-4',
```
adatzer: nit - would it make sense to move the schema strings (for `payload_data`, `deep_link_received`, `message_notification`, `timing`) e.g. to `constants.ts`? Not an objection at all though, just for better visibility.

src/jsCore.ts
```diff
@@ -0,0 +1,509 @@
...
+      schema:
+        'iglu:com.snowplowanalytics.mobile/message_notification/jsonschema/1-0-0',
+      data: details.eventData,
+    },
+    contexts: details.contexts,
+  });
+}
+
+function removeGlobalContexts(): Promise<void> {
+  return <Promise<void>>Promise.reject('Not implemented');
```
adatzer: for consistency with the other rejects (e.g. [here](https://github.com/snowplow/snowplow-react-native-tracker/blob/20f6000c6e37ae903abb75d7b421a373ef207f65/src/events.ts#L58)), we could also make the reason an instance of error

src/jsCore.ts
matus-tomlein: Ohh, I see, that's true, setting `null` in the native tracker resets the property but here it is ignored. Would be better if it behaved the same way, but unfortunately the JS tracker core doesn't accept `null` and doesn't provide a way to reset the values. We could do something custom in this jsCore wrapper but I would prefer keep it very simple at this time.

.github/workflows/deploy.yml
matus-tomlein: Good catch! Should have tried the command before committing it 🙈 Have tried your fix and that works, thanks!

src/jsCore.ts
colmsnowplow: AFAIR if you want to unset userid in JS tracker you pass an empty string. Bit odd/jank, probably should change, but that's the behaviour (at least last I checked, which is a while ago) fwiw

(Sorry to barge in :D )

matus-tomlein: Good shout, Colm! I think you are right, at least [this check here](https://github.com/snowplow/snowplow-javascript-tracker/blob/50fb7ea55b1bc7d41376537bfb78d512fc2e7683/libraries/tracker-core/src/payload.ts#L121) suggests that empty strings are ignored when adding to payload. Will change it to set empty strings, thanks!

.github/workflows/e2e-android.yml
matus-tomlein: Good point, updated!

src/native.ts
matus-tomlein: Moved it to `types.ts` 👍

src/jsCore.ts
matus-tomlein: Good idea, added the freeze

src/jsCore.ts
matus-tomlein: Moved to constants

src/jsCore.ts
matus-tomlein: Updated, thanks!

src/jsCore.ts
```diff
@@ -0,0 +1,497 @@
...
+  return <Promise<void>>Promise.resolve();
+}
+
+function trackScreenViewEvent(details: {
+  tracker: string | null;
+  eventData: ScreenViewProps;
+  contexts: EventContext[];
+}): Promise<void> {
+  forTracker(details.tracker, (tracker) => {
+    tracker.track(buildScreenView(details.eventData), details.contexts);
```
adatzer: Just noticed that [the screen view schema used in native](https://github.com/snowplow/snowplow-objc-tracker/blob/ec912a566073615ebcc9dc3a4c6b96a1cc6ed7e7/Snowplow/Internal/SPTrackerConstants.m#L57) is different than the one used by [`buildScreenView`](https://github.com/snowplow/snowplow-javascript-tracker/blob/4dec01b9a314ca11d88372afa4b2dc67a431ee0d/libraries/tracker-core/src/core.ts#L814) in JavaScript tracker. I did not know that! This means we'd need to switch to `trackSelfDescribingEvent`?

src/jsCore.ts
```diff
@@ -0,0 +1,497 @@
...
+      version: details.eventData.version,
+      name: details.eventData.name,
+      description: details.eventData.documentDescription,
+    });
+    tracker.track(built.event, details.contexts.concat(built.context));
+  });
+  return <Promise<void>>Promise.resolve();
+}
+
+function trackEcommerceTransactionEvent(details: {
```
adatzer, line 323: Not sure about this, but are the mobile trackers tracking `transaction_item` events automatically besides the `transaction` event? Not sure if we'd like parity here. What do you think?

src/jsCore.ts
matus-tomlein: Life-saving catch again, thanks Ada! I completely missed that the events are tracked with the other screen view schema. Fixed it.

src/jsCore.ts
matus-tomlein: You're right, the mobile trackers track the items as separate events. Change it here as well. Great spot, thanks!

adatzer (approved):
This is great! Very happy to approve!

PR #200, opened 2023-09-06
Upgrade project structure and update native trackers to version 5 (close #199)

### LINKED ISSUE #199: Update native trackers to version 5
Currently both native trackers are checked out on version 4. Are there any plans to update to version 5?

matus-tomlein:
In addition to upgrading the native trackers to v5, this PR brings bigger changes in the project which are aimed to make it easier to maintain it in the future.

## Project structure

I ran into a series of problems when trying to update the project and demo app dependencies which ultimately made me choose to recreate the project from a blank template (using [create-react-native-library scaffold](https://github.com/callstack/react-native-builder-bob)).

I decided to do that because the way that we built the demo apps before was quite fragile. First, we built the main project using `npm` and the demo apps using `yarn`. The demo apps included the main project using a `file:..` reference which copied the whole `node_modules` from the parent project that we then had to fix in a `postinstall` hook.

Instead, the new structure uses `yarn` for the whole project. All the build commands are executed from the top level (root folder) without switching into the example app sub-folder. In case one wants to execute a script that is defined in the example project (e.g., `android` to start the app on Android emulator), they can still do it like this:

```sh
yarn example android
```

I think that this simplifies the project and will make it easier to maintain and update in the future. There are also various other improvements that make it easier to work with the project such running `yarn` also installs the pods.

## Translated Objective-C to Swift and Java to Kotlin

The second reason why I thought it's better to recreate the project is that this allowed me to adopt Swift and Kotlin (instead of Objective-C and Java) for the native code. Since the native trackers version 5 also transition to Swift and Kotlin, I thought it'd be better to make the change here too. This should also make it much easier to maintain the project and add new features.

## Removed TV demo app

Another change aimed to improve the maintainability was to remove the TV demo app from the project. The TV demo app was exactly the same code as the other demo, it just had a different target to run on tvOS and Android TV. Having the same project duplicated twice was a burden for us to maintain, keep the dependencies up to date in both places and test. I don't think it provided enough value for us to justify it. We may decide to put the TV demo app in a separate repository with other examples (similar like we do for the JavaScript tracker).

## Upgraded mobile native trackers to version 5

Finally, this PR upgrades the underlying native trackers to version 5, addressing issue #199.

One side-effect of upgrading to version 5 is that now it is necessary for users to add `FMDB` to their `ios/Podfile` file (unless using Expo Go). They will basically need to add this line to the file:

```rb
pod 'FMDB', :modular_headers => true
```

The reason for it is that the new iOS tracker is pure Swift and pure Swift projects in CocoaPods require that all dependencies have modular imports. However, the iOS tracker itself can't make define that, so we need to do it in the user app. I will make sure to document this in the install guide.

## Reviewing

I appreciate that this is a huge change across many files that is almost impossible to review thoroughly. I tagged multiple reviewers because I thought it might be better to have more people do a light review to make it more likely to spot any problems. Thanks and please let me know if there is any way I can help with making it easier to review!


.github/workflows/e2e-ios.yml
```diff
@@ -6,132 +6,85 @@ on:
 jobs:
   e2e:
     runs-on: macos-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
```
adatzer, line 9: not an issue, just wondering whether we do need such a long timeout

.github/workflows/deploy.yml
```diff
@@ -1,59 +1,36 @@
...
-        IOS_VER_FILEPATH: 'ios/Util/RNTrackerVersion.m'
-        ANDR_VER_FILEPATH: 'android/src/main/java/com/snowplowanalytics/react/util/TrackerVersion.java'
+        IOS_VER_FILEPATH: 'ios/Util/TrackerVersion.swift'
+        ANDR_VER_FILEPATH: 'android/src/main/java/com/snowplow/reactnativetracker/util/TrackerVersion.kt'
         JS_VER_FILEPATH: 'src/jsCore.ts'
       run: |
         echo ::set-output name=TAG_VERSION::${GITHUB_REF#refs/*/}
         echo "##[set-output name=RN_TRACKER_VERSION;]$(node -p "require('./package.json').version")"
-        echo "##[set-output name=RN_IOS_TRACKER_VERSION;]$(cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = @"rn-\(.*\)";/\1/p')"
+        echo "##[set-output name=RN_IOS_TRACKER_VERSION;]$(cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = @"rn-\(.*\)"/\1/p')"
```
adatzer, line 33: `@` needs to be removed from `sed`, i.e. to change to 

```
cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = "rn-\(.*\)"/\1/p'
```

.eslintrc.js
```diff
@@ -1,76 +0,0 @@
-module.exports = {
```
adatzer, line 1: what is this file been replaced with? i see the eslint section in package.json but is not the same (e.g. promises) - would we want to keep the same lint rules?

mscwilson (commented):
On iOS, there's a problem with lifecycle tracking. I'm getting background events but no foreground, and the printed out session info says `"backgroundIndex":0,"foregroundIndex":0`.

Nit: for anonymous tracking, the `previousSessionId` is null, I was expecting it to be all 0s like the `userId`

Nit: some of the UUIDs are coming up capitalised on iOS. Is that normal for iOS tracker?

anon_session

matus-tomlein:
Thanks a lot @mscwilson!

Regarding the missing foreground event, I think that is a problem with the underlying iOS tracker. I know it does not happen in the iOS tracker demo app, but my guess is that's just a coincidence. Everything that is relevant to the foreground/background event tracking is handled by the iOS tracker. I could reproduce what happens in the iOS tracker and have created a task for it there including a hypothesis that I have: https://github.com/snowplow/snowplow-ios-tracker/issues/818

The `userId` is not null like the `previousSessionId` during anonymous tracking because it doesn't accept null in the schema. Otherwise we would make both null, I think that makes more sense than a 0-filled UUID.

I don't know where the difference in casing in the UUIDs is coming from, but that is just a visual difference because UUID is case insensitive.

mscwilson (approved):
LGTM! This is a big improvement for this tracker.

.github/workflows/deploy.yml
```diff
@@ -1,59 +1,36 @@
 name: Deploy
 
 on:
-  push:
-    tags:
-    - '*.*.*'
+  workflow_run:
+    workflows: [Build]
+    branches: [main]
```
adatzer, line 6: Our default is `master` not `main`.

Also does that mean that we cannot run a deploy run from other branches (e.g. for `rc` releases)?

.github/workflows/deploy.yml
```diff
@@ -1,59 +1,36 @@
...
-        ANDR_VER_FILEPATH: 'android/src/main/java/com/snowplowanalytics/react/util/TrackerVersion.java'
+        IOS_VER_FILEPATH: 'ios/Util/TrackerVersion.swift'
+        ANDR_VER_FILEPATH: 'android/src/main/java/com/snowplow/reactnativetracker/util/TrackerVersion.kt'
         JS_VER_FILEPATH: 'src/jsCore.ts'
       run: |
         echo ::set-output name=TAG_VERSION::${GITHUB_REF#refs/*/}
         echo "##[set-output name=RN_TRACKER_VERSION;]$(node -p "require('./package.json').version")"
-        echo "##[set-output name=RN_IOS_TRACKER_VERSION;]$(cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = @"rn-\(.*\)";/\1/p')"
+        echo "##[set-output name=RN_IOS_TRACKER_VERSION;]$(cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = @"rn-\(.*\)"/\1/p')"
         echo "##[set-output name=RN_ANDROID_TRACKER_VERSION;]$(cat "${ANDR_VER_FILEPATH}" | sed -n -e 's/^.*RN_TRACKER_VERSION = "rn-\(.*\)";/\1/p')"
```
adatzer, line 34: Semicolon needs to be removed from sed as well, i.e. to:

```
cat "${ANDR_VER_FILEPATH}" | sed -n -e 's/^.*RN_TRACKER_VERSION = "rn-\(.*\)"/\1/p'
```

.github/workflows/deploy.yml
```diff
@@ -1,59 +1,36 @@
...
         echo "##[set-output name=RN_JS_TRACKER_VERSION;]$(cat "${JS_VER_FILEPATH}" | sed -n -e 's/^.*trackerVersion = '\''rn-\(.*\)'\'';/\1/p')"
```
adatzer, line 35: Nit, but it's a nice opportunity to remove warning for deprecated `set-output`. Since we later reference the `version.outputs`, we can set them as [output parameters](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)?

README.md
```diff
@@ -9,28 +8,31 @@
...
 
 With this library you can collect granular event-level data as your users interact with your React Native applications. It is build on top of Snowplow's [Mobile Native][native-trackers] [iOS][objc-tracker] and [Android][android-tracker] Trackers, in order to support the full range of out-of-the-box Snowplow events and tracking capabilities.
 
-
 ## Quick start
 
 From the root of your [React Native][react-native] project:
 
-```
+```sh
```
adatzer, line 25: do these commands need updating as well?

android/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,3 @@
-
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="com.snowplowanalytics.react.tracker">
-
+          package="com.snowplow.reactnativetracker">
```
adatzer, line 2: apologies for ignorance, why is this change needed and what is its effect?

example/src/App.js
```diff
@@ -31,14 +31,14 @@ import {
 /**
  * URI of the Snowplow collector (e.g., Micro, Mini, or BDP) to send events to
  */
-const collectorEndpoint = 'placeholder';
+const collectorEndpoint = 'http://192.168.100.2:9090';
```
adatzer, line 34: nit - leave as placeholder?

src/__mocks__/react-native.js
```diff
@@ -4,7 +4,7 @@
 module.exports = {
   NativeModules: {
     RNSnowplowTracker: {
-      createTracker: (n,cc) => Promise.resolve(true)
-    }
-  }
+      createTracker: (n, cc) => Promise.resolve(true),
```
adatzer, line 7: based on the check warnings, it seems we can remove the `eslint-disable` comments in this file

snowplow-react-native-tracker.podspec
```diff
@@ -0,0 +1,42 @@
+require "json"
+
+package = JSON.parse(File.read(File.join(__dir__, "package.json")))
+folly_compiler_flags = '-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1 -Wno-comma -Wno-shorten-64-to-32'
+
+Pod::Spec.new do |s|
+  s.name         = "snowplow-react-native-tracker"
```
adatzer, line 7: the name was `RNSnowplowTracker` before, i don't mind changing it, however i am not sure what are the consequences, if any

src/__tests__/events.test.ts
```diff
@@ -165,24 +189,30 @@ describe('test validateStructured', () => {
...
+    const testEv = { pageTitle: 'test' } as any;
+    await expect(e.validatePageView(testEv)).rejects.toThrow(
+      logMessages.pageviewReq
+    );
   });
 
   test('test invalid page-view - missing pageUrl', async () => {
-    const testEv = {pageUrl: null} as any;
-    await expect(e.validatePageView(testEv)).rejects.toThrow(logMessages.pageviewReq);
+    const testEv = { pageUrl: null } as any;
```
adatzer, line 205: nit to rename test, only because of the check warning for this test:

```
Test title is used multiple times in the same describe block
```

adatzer (commented):
This is awesome work @matus-tomlein !
I see this PR has been merged into `release/2.0.0` already, so feel free to ignore the comments below until the release PR.

README.md
matus-tomlein: No, these should be okay.

.github/workflows/e2e-ios.yml
matus-tomlein: On the first run, when there is nothing in the cache, it does take close to an hour to complete the e2e tests on iOS.

.eslintrc.js
matus-tomlein: I am keen to go with the defaults in terms of linting and formatting. This PR added `prettier` as the formatter so I think we should be covered well.

android/src/main/AndroidManifest.xml
matus-tomlein: It is not needed, but it was caused by recreating the project with RN defaults – based on the project name, the generator went for this. It doesn't really matter because it's internal within the package, does not have any effect outside of tracker internals.

snowplow-react-native-tracker.podspec
matus-tomlein: This is also just an internal naming within the library, not a published name, so should have no external effects.

PR #201, opened 2023-09-07
Fix buffer option in emitter configuration on Android to set correct value for large setting (close #197)

### LINKED ISSUE #197: EmitterConfig bufferOption does not match types on Android
**Describe the bug**
The largest value for the `bufferOption` in the `emitterConfig` is typed as `large` for both iOS and Android, but it should be `large` for iOS and `heavy` for Android

**To Reproduce**
Compare [iOS](https://github.com/snowplow/snowplow-react-native-tracker/blob/c151a65e3b37d40b0c09052ad0727212428d36df/ios/Util/RNConfigUtils.m#L92)/[Android](https://github.com/snowplow/snowplow-react-native-tracker/blob/20f6000c6e37ae903abb75d7b421a373ef207f65/android/src/main/java/com/snowplowanalytics/react/util/ConfigUtil.java#L84) -> iOS converts `large` to the largest buffer option. If `large` is given to Android; it uses `single`.

**Expected behavior**
The interface should be standardized (e.g. `large` should map to the biggest bufferOption on both platforms?)

**Screenshots**

 - Device: -
 - OS: -
 - Browser: -
 - Version: snowplow-react-native-tracker 1.4.0

**Additional context**

matus-tomlein:
Fixes issue #197 in which the `large` buffer option setting was incorrectly parsed on Android (it was expected to be called `heavy` instead of `large`). Also does some small refactoring in the file along the way.

### DIFF
--- a/android/src/main/java/com/snowplow/reactnativetracker/util/ConfigUtil.kt
+++ b/android/src/main/java/com/snowplow/reactnativetracker/util/ConfigUtil.kt
@@ -15,68 +15,6 @@ import com.snowplowanalytics.snowplow.util.TimeMeasure
 import java.util.concurrent.TimeUnit
 
 object ConfigUtil {
-  fun mkDevicePlatform(devPlatform: String): DevicePlatform {
-    var devicePlatform = DevicePlatform.Mobile
-    if (devPlatform == "web") {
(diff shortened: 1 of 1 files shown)


mscwilson (approved):
lgtm, nice improvement to the code as well

PR #202, opened 2023-09-07
Disable automatic screen view tracking in the iOS and Android tracker by default (close #198)

### LINKED ISSUE #198: Disable automatic screen view tracking in the iOS and Android tracker that is now enabled by default
The automatic screen view tracking functionality of the underlying iOS and Android trackers produces not useful events on React Native – it just tracks `RNScreen` and similar events that don't mean much.

We should disable it by default and guide users to implement screen view autotracking using the React Native navigation frameworsk.

matus-tomlein:
The automatic screen view tracking functionality of the underlying iOS and Android trackers produces not useful events on React Native – it just tracks RNScreen and similar events that don't mean much.

This PR disables the feature by default.


src/types.ts
```diff
@@ -170,8 +170,10 @@ export interface TrackerConfiguration {
    */
   screenContext?: boolean;
   /**
-   * Whether enable automatic tracking of ScreenView events.
-   * @defaultValue true
+   * Whether enable automatic tracking of ScreenView events from the native side.
```
mscwilson: nit
```suggestion
   * Whether to enable automatic tracking of ScreenView events from the native side.
```

mscwilson (approved):
lgtm

PR #203, opened 2023-09-29
Release/2.0.0

### LINKED ISSUE #197: EmitterConfig bufferOption does not match types on Android

### LINKED ISSUE #198: Disable automatic screen view tracking in the iOS and Android tracker that is now enabled by default

### LINKED ISSUE #199: Update native trackers to version 5

matus-tomlein:
This major release brings internal changes in the project structure that are aimed to make it easier to maintain it going forward. It also updates the underlying iOS and Android trackers to the latest major version 5 and adopts Swift and Kotlin (instead of Objective-C and Java) for the all the native code, making it safer and more modern.

## Changelog

* Upgrade project structure and update native trackers to version 5 (#199)
* Disable automatic screen view tracking in the iOS and Android tracker by default (#198)
* Fix buffer option in emitter configuration on Android to set correct value for large setting (#197)

[PR for documentation updates is here](https://github.com/snowplow/documentation/pull/621).


mscwilson (approved):
LGTM! great update

.github/workflows/deploy.yml
```diff
@@ -1,60 +1,37 @@
 name: Deploy
 
 on:
-  push:
-    tags:
-    - '*.*.*'
+  workflow_run:
+    workflows: [Build]
+    branches: [master]
```
adatzer: if `deploy` runs only on `master`, how could we release an `rc` version?

.github/workflows/deploy.yml
```diff
@@ -1,60 +1,37 @@
...
+        IOS_VER_FILEPATH: 'ios/Util/TrackerVersion.swift'
+        ANDR_VER_FILEPATH: 'android/src/main/java/com/snowplow/reactnativetracker/util/TrackerVersion.kt'
         JS_VER_FILEPATH: 'src/jsCore.ts'
       run: |
         echo ::set-output name=TAG_VERSION::${GITHUB_REF#refs/*/}
         echo "##[set-output name=RN_TRACKER_VERSION;]$(node -p "require('./package.json').version")"
-        echo "##[set-output name=RN_IOS_TRACKER_VERSION;]$(cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = @"rn-\(.*\)";/\1/p')"
-        echo "##[set-output name=RN_ANDROID_TRACKER_VERSION;]$(cat "${ANDR_VER_FILEPATH}" | sed -n -e 's/^.*RN_TRACKER_VERSION = "rn-\(.*\)";/\1/p')"
+        echo "##[set-output name=RN_IOS_TRACKER_VERSION;]$(cat "${IOS_VER_FILEPATH}" | sed -n -e 's/^.*kRNTrackerVersion = "rn-\(.*\)"/\1/p')"
+        echo "##[set-output name=RN_ANDROID_TRACKER_VERSION;]$(cat "${ANDR_VER_FILEPATH}" | sed -n -e 's/^.*RN_TRACKER_VERSION = "rn-\(.*\)"/\1/p')"
```
adatzer: nit - could use [output parameters](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter) instead of `set-output` that [is deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

adatzer (approved):
Looks great!!

.github/workflows/deploy.yml
matus-tomlein: Good point, have added the `next` branch to the list of branches there too.

.github/workflows/deploy.yml
matus-tomlein: Have replaced the `set-output`.

PR #208, opened 2024-02-21
Release/2.1.0

### LINKED ISSUE #90: Allow for a 'page ping'-like event for screen views (screen-ping?)
I would like to be able to reliably calculate how long a user stays active on a screen in a mobile app - it can be difficult to handle events like a minimized app or screen off events.

An event analogous to the page_ping in the web data model would be useful - a `screen_ping` event, with a user-configurable interval?

For now, I am using the time between observed screen views, with some edge-case-handling involving app minimization events, in tandem with parsing intervals between screen views tha feel "too long" - but of course this is somewhat arbitrary.

### LINKED ISSUE #204: ReactNativeTracker.trackSelfDescribingEvent typing issue
**Describe the bug**
There seems to be an issue at the `ReactNativeTracker.trackSelfDescribingEvent` API.

image

MediaPlayer being:
```ts
interface MediaPlayer {
    /**
     * The current playback time position within the media in seconds.
     */
    currentTime: number;
    /**
     * A floating-point value indicating the duration of the media in seconds.
     */
    duration?: number | null;
    /**
     * Whether playback of the media has ended.
     */
    ended: boolean;
    /**
     * Whether the video element is fullscreen.
     */
    fullscreen?: boolean | null;
    /**
     * Human readable name given to tracked media content.
     */
    label?: null | string;
    /**
     * Whether the media is a live stream.
     */
    livestream?: boolean | null;
    /**
     * Whether the video should restart after ending.
     */
    loop?: boolean | null;
    /**
     * Type of media content.
     */
    mediaType?: MediaType | null;
    /**
     * Whether the media element is muted.
     */
    muted?: boolean | null;
    /**
     * Whether the media element is paused
     */
    paused: boolean;
    /**
     * Whether the video element is showing picture-in-picture.
     */
    pictureInPicture?: boolean | null;
    /**
     * Playback rate (1 is normal).
     */
    playbackRate?: number | null;
    /**
     * Type of the media player (e.g., com.youtube-youtube, com.vimeo-vimeo,
     * org.whatwg-media_element).
     */
    playerType?: null | string;
    /**
     * Quality level of the playback (e.g., 1080p).
     */
    quality?: null | string;
    /**
     * Volume percentage at which the media will be played.
     */
    volume?: number | null;
}
```

The issue seems to be coming from this line: `extends Record<keyof TData, unknown>` on the type definition of `trackSelfDescribingEvent`.

**Expected behavior**
Types work properly with any valid interface for a schema.

matus-tomlein:
This release updates the underlying mobile trackers to version 6 and adds mobile screen engagement tracking. It also fixes the type for the self-describing event data.

**Enhancements**
Upgrade to mobile trackers version 6 and add screen engagement tracking (#90)

**Bug fixes**
Fix SelfDescribing extension type (#204)

**Under the hood**
Update JavaScript tracker core version to 3.21.0
Fix running deploy action only on new tags
Update cocoapods version
Lock yarn version to 1.22.19
Update dev dependencies


igneel64 (approved):
Seems ok for the typing part!

CHANGELOG
```diff
@@ -1,3 +1,13 @@
+Version 2.1.0 (2023-02-21)
```
greg-el: 2023? 👀 
```suggestion
Version 2.1.0 (2024-02-21)
```

src/__tests__/events.test.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2023 Snowplow Analytics Ltd. All rights reserved.
+ * Copyright (c) 2020-present Snowplow Analytics Ltd. All rights reserved.
```
greg-el: This applies to all of these changes - I think we've just been removing the licenses in files in the JS tracker completely in favour of just the LICENSE file, perhaps we could do the same throughout this also?

src/api.ts
```diff
@@ -109,6 +111,36 @@ function trackScreenViewEvent(namespace: string) {
   };
 }
 
+/**
+ * Returns a function to track a ScrollChanged event by a tracker
+ *
+ * @param namespace {string} - The tracker namespace
+ * @returns - A function to track a ScrollChanged event
+ */
+function trackScrollChangedEvent(namespace: string) {
```
greg-el, line 107: Could a return type be added here? (and the function below)

greg-el (approved):
A few comments but otherwise looks good to me!

src/api.ts
matus-tomlein: Good catch, but the type is defined implicitly by the function on the line below, not sure if it's worth the extra trouble to define it here as well – TypeScript is pretty good at the type inference.

src/__tests__/events.test.ts
matus-tomlein: Good call, have now removed the license from the source files!

PR #213, opened 2024-10-29
Fix crash on iOS on React Native new architecture (close #212)

### LINKED ISSUE #212: [Fabric] - Snowplow crashes with new architecture enabled
**Describe the bug**
Immediate crash when `createTracker` is called with new architecture enabled

**To Reproduce**
```typescript
import { createTracker, ReactNativeTracker } from '@snowplow/react-native-tracker'

export const App: React.FC = () => {
 createTracker(...)
 return null
}
```

**Expected behavior**
- Should not crash with Fabric enabled

**Screenshots**

 - Device: iPhone 15 Pro
 - OS: 17.5.1
 - Version `"@snowplow/react-native-tracker": "^2.1.1"`

matus-tomlein:
Issue #212 

The culprit why the tracker crashed on iOS under the RN new architecture was [this extra ObjC function](https://github.com/snowplow/snowplow-react-native-tracker/pull/213/files#diff-d942ab96abb752524b9bf52f3ede887e6f8a0b4bdd8088ddf87bc170e185c0bdL5) in the code – it shouldn't have been there and was missing implementation. So the fix was just removing that function.

I have tested this separately in a React Native app with new architecture enabled. I wanted to also add the new architecture to CI but that will require a separate PR since it completely breaks the e2e and build actions 😅

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -111,7 +111,7 @@ jobs:
 
       - name: Upload events artifact if Detox-Jest failed
         if: ${{ failure() && steps.dj.outcome == 'failure' && steps.eventlog.outcome == 'success' }}
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         with:
           name: android-events-artifact
           path: android-events-micro.json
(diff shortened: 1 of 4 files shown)


mscwilson (approved):
lgtm

PR #215, opened 2024-10-30
Release/2.1.2

matus-tomlein:
This patch release adds support for React Native new architecture by fixing an issue on iOS that caused crashes when the tracker was used under the new architecture.

**Bug fixes:**

- Fix crash on iOS on React Native new architecture (#212)


mscwilson (approved):
how satisfying that issue 212 is for v2.1.2.
lgtm!

